PR #1, opened 2022-03-07
Create jwt with pyjwt

gowoons:
## Purpose

Change jwt management by using pyjwt directly instead of rest simplejwt module. This allows better modularity and simplicity.

## Proposal

- [x] Use jwt package instead of rest_framework_simplejwt
- [x] Generate access token
- [x] Generate refresh token
- [x] Save refresh token to session
- [x] Safely redirect to jitsi with jwt token
- [x] Implement tests


src/backend/jitsi_magnify/core/api.py
```diff
@@ -32,42 +37,73 @@ def exception_handler(exc, context):
...
+    encoded_params = urlencode(url_params)
+    url = f"https://{base_url}/{room}?{encoded_params}"
+
+    # As we allow users to pass information to redirect url (with the parameter room),
+    # we check that this url is safe before redirecting
+    if not is_safe_url(url, allowed_hosts=base_url, require_https=True):
+        return Response(f"Redirection to url {url} is not allowed", status=404)
+
+    response = HttpResponseRedirect(url)
+    response.set_cookie(
```
sampaccoud, line 102: not sure how this cookie is used?
I found another dependency to simple jwt here https://github.com/openfun/jitsi-magnify/blob/main/src/backend/jitsi_magnify/settings.py#L183 :wink:

src/backend/jitsi_magnify/core/api.py
```diff
@@ -32,42 +37,73 @@ def exception_handler(exc, context):
...
 
-    token = CustomTokenSerializer.get_token(user=user, room=room)
-
-    return HttpResponseRedirect(settings.JWT_CONFIGURATION.get("jitsi_url") + "/" + room + "?jwt=" + str(token))
+    access_token_expiration = datetime.datetime.utcnow() + datetime.timedelta(minutes=5)
+    refresh_token_expiration = datetime.datetime.utcnow() + datetime.timedelta(days=7)
+    access_token = generate_token(user, room, access_token_expiration)
+    refresh_token = generate_token(user, room, refresh_token_expiration)
+
+    base_url = settings.JWT_CONFIGURATION.get("jitsi_domain_name")
```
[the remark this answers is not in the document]
gowoons: I understand, but what do you mean exactly by "making this setting required" ? Making the app exit with an error if the setting does not exist ?

src/backend/jitsi_magnify/core/api.py
gowoons: It is not used yet, but tokens can usually be refreshed with a route such as /api/token/refresh, with the refresh token that is kept in local storage (here in cookies). I am still documenting myself on this but it should be implemented in the future

src/backend/jitsi_magnify/core/api.py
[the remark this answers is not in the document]
sampaccoud: By not setting any default value in the settings file:
```
"jitsi_domain": values.Value(
    environ_name="JWT_JITSI_DOMAIN", environ_prefix=None
)
```

src/backend/jitsi_magnify/core/api.py
sampaccoud: Well... actually we don't need that because we will generate a fresh JWT token for you anytime you call this view... refreshing the token is required when it is used as the main authentication. Right?
If you agree, then we can remove this?

gowoons: It should not be required here, yes, so I will remove everything related to the refresh token. We'll have to test once jitsi-magnify is deployed that there are no problems once access_token is expired (such as accessing other participants' videos, muting, etc...), however we think that as soon as it connects to the web socket for jitsi, it does not need the token anymore

README.md
```diff
@@ -67,7 +67,13 @@ $ make superuser
...
 ### Explanation
 
-# @TODO add explanation on feature
+Jitsi-magnify helps with authenticating users when they access jitsi rooms. Currently, magnify creates a guest user and generates a token that gives access to the room when contacted on route /api/token/{room}.
+
+### Usage in production
+
+The domain of jitsi instance that uses magnify should be referenced in env.d/development/common. As for jitsi, env variables should be set to enable jwt auth and to redirect to magnify instance. 
+
+These variables are : `ENABLE_AUTH=1`, `AUTH_TYPE=jwt`, `JWT_APP_ID=<app_id>`(equal to JWT_JITSI_APP_ID here), `JWT_APP_SECRET=<app_secret>`(equal to JWT_JITSI_PRIVATE_SIGNING_KEY) and `TOKEN_AUTH_URL=https://JWT_JITSI_DOMAIN/api/token/{room}`.
```
sampaccoud, line 76: It would be clearer if you make it several lines?

src/backend/jitsi_magnify/settings.py
```diff
@@ -179,38 +178,25 @@ class Base(Configuration):
...
-        "jitsi_url": values.Value(
-            "", environ_name="JWT_JITSI_URL", environ_prefix=None
+        "jitsi_domain": values.Value(
+            environ_name="JWT_JITSI_DOMAIN", environ_prefix=None
         ),
         "jitsi_app_id": values.Value(
             "", environ_name="JWT_JITSI_APP_ID", environ_prefix=None
         ),
+        "jitsi_secret_key": values.SecretValue(
+            environ_name="JWT_JITSI_PRIVATE_SIGNING_KEY", environ_prefix=None
```
sampaccoud, line 198: Maybe uses the same name?
```suggestion
            environ_name="JWT_JITSI_SECRET_KEY", environ_prefix=None
```

src/backend/jitsi_magnify/core/api.py
```diff
@@ -32,42 +37,63 @@ def exception_handler(exc, context):
...
-class CustomTokenView(TokenObtainPairView):
-    """View for token creation"""
-    serializer_class = CustomTokenSerializer
 
-def obtainTokenView(request, room):
+@api_view(("GET",))
+def get_room_token_view(request, room):
     """View for token fetching based on a guest user and a specific room
     This redirects to jitsi with the token"""
     user, created = User.objects.get_or_create(username="guest")
```
sampaccoud, line 81: I think we should not create a user in database for a guest. This code made sense in joanie because al users are connected. Here, you can check in `generate_token` if `request.user.is_authenticated()` and use a generic user id, username and no email?

src/backend/jitsi_magnify/core/api.py
```diff
@@ -32,42 +37,63 @@ def exception_handler(exc, context):
...
-    
     if created:
         user.set_password(settings.JWT_CONFIGURATION.get("guest_default_password"))
         user.save()
```
sampaccoud, line 84: see above comment, so this can also be removed.

src/backend/jitsi_magnify/core/api.py
```diff
@@ -32,42 +37,63 @@ def exception_handler(exc, context):
...
 
-    token = CustomTokenSerializer.get_token(user=user, room=room)
+    access_token_expiration = datetime.datetime.utcnow() + datetime.timedelta(minutes=5)
```
sampaccoud, line 86: You should maybe make a setting for token validity instead of hardcoding 5?

src/backend/jitsi_magnify/core/api.py
```diff
@@ -32,42 +37,63 @@ def exception_handler(exc, context):
...
+    access_token = generate_token(user, room, access_token_expiration)
```
sampaccoud, line 87: I don't see the point to pass expiration as a parameter. Better compute expiration directly in the function?

sampaccoud:
You could add a simple test by doing that makes a GET on your endpoint and checks that the redirect is made as expected.
See https://docs.djangoproject.com/en/2.0/topics/testing/tools/#django.test.SimpleTestCase.assertRedirects

src/backend/jitsi_magnify/core/api.py
gowoons: It was useful when there was also a refresh token that had a different expiration, but yes now it's not needed

src/backend/jitsi_magnify/core/api.py
```diff
@@ -32,42 +37,69 @@ def exception_handler(exc, context):
...
 
-    @classmethod
-    def get_token(cls, user, room):
-        """Method to get the token created based on a user and a room which they wish to access"""
-        token = super().get_token(user)
+def create_payload(user, room):
+    """Create the payload so that it contains each information jitsi requires"""
+    expiration_minutes = settings.JWT_CONFIGURATION.get("token_expiration_minutes")
+    expiration_hours = settings.JWT_CONFIGURATION.get("token_expiration_hours")
+    expiration_days = settings.JWT_CONFIGURATION.get("token_expiration_days")
```
sampaccoud, line 45: Actually, you can just use seconds to ease configuration. It does not prevent us from setting expiration to 259200 seconds if we want 3 days :man_shrugging:

src/backend/jitsi_magnify/core/api.py
```diff
@@ -32,42 +35,103 @@ def exception_handler(exc, context):
...
+        "aud": "jitsi",
+        "iss": settings.JWT_CONFIGURATION.get("jitsi_app_id"),
+        "sub": "meet.jitsi",
+        "room": room,
+    }
 
-        token['context'] = {'user': jitsiUser}
+    jitsi_user = {
+        "avatar": settings.JWT_CONFIGURATION.get("guest_avatar"),
+        "name": user.username,
```
sampaccoud, line 57: @thomas-borot I saw your commit. But it could just be done like this maybe?
```suggestion
        "name": user.username if user.is_authenticated else settings.JWT_CONFIGURATION.get("guest_name"),
```

README.md
```diff
@@ -67,7 +67,21 @@ $ make superuser
 
 ### Explanation
 
-# @TODO add explanation on feature
+Jitsi-magnify helps with authenticating users when they access jitsi rooms. Currently, magnify creates a guest user and generates a token that gives access to the room when contacted on route /api/token/{room}.
```
sampaccoud, line 70: > magnify creates a guest user

Not any more :wink:

src/backend/jitsi_magnify/tests/test_redirect.py
```diff
@@ -0,0 +1,63 @@
...
+import jwt
+
+
+class RedirectTestCase(TestCase):
+    """Test suite for redirection to jitsi instance"""
+
+    def test_redirection(self):
+        """
+        Get path /api/token/test and redirect to jitsi with a token
+        We thus verify that this redirects as wanted, and that the token is correct
```
sampaccoud, line 20: The test is nice but philosophically not 100% satisfying.

You are placing yourself in the context of your dev environment, at the time it is now, and then you are recalculating the token with the same method as the code you are trying to test... This means that for example, if the underlying jwt library changes its behavior, you could miss a big change that could potentially break the feature without you noticing.

A better test philosophy is to define the whole context for your test and then check that the token is exactly what you expect. It's also much easier to write such test than recalculating everything like you do :wink: 

In this case you would: 
- define a test value for all settings impacting your tests thanks to the `override_settings` decorator
- choose a fix date to mock "timezone.now", not the moment the test is running like you do here
- hardcode the complete url you expect, including the encoded token and check that the request is redirected to it when you call the url.

This way you will be sure that the url works exactly how you expect and returns the correct value.

Also, you should write another test for a logged in user (self.client.login(...))

src/backend/jitsi_magnify/core/views.py
```diff
@@ -1 +1,71 @@
...
+from django.utils.http import url_has_allowed_host_and_scheme
+
+import jwt
+
+
+def create_payload(user, room):
+    """Create the payload so that it contains each information jitsi requires"""
+    expiration_seconds = int(settings.JWT_CONFIGURATION.get("token_expiration_seconds"))
+    print(user)
+    print(user.is_authenticated)
```
sampaccoud, line 17: :hand_over_mouth:

src/backend/jitsi_magnify/core/views.py
```diff
@@ -1 +1,71 @@
...
+    jitsi_user = {
+        "avatar": settings.JWT_CONFIGURATION.get("guest_avatar"),
+        "name": user.username
+        if user.is_authenticated
+        else settings.JWT_CONFIGURATION.get("guest_username"),
+        "email": user.email if user.is_authenticated else "",
+    }
+
+    token_payload["context"] = {"user": jitsi_user}
+    print(token_payload)
```
sampaccoud, line 37: :zipper_mouth_face:

sampaccoud (approved):

sampaccoud (approved):

PR #7, opened 2022-06-09
🎉(root) Create the project structure [WIP]

Leonils:
## Purpose

Create a 4-part structure, (frontend + backend) * (lib + demo), implementing the following tools and features:

## Proposal

This PR should provide following features:

- We can start the demo backend
- The magnify folder is a package exporting a default set of settings and a single app "rooms"
- We can work on the rooms app and it updates on the demo backend instantly
- We can test rooms by writing tests in test/apps/rooms
- We can write react components in the folder frontend
- We can see them in isolation in a storybook
- We can test them using testing libraries
- This frontend library is imported from the frontend demo, also up and running as "@jitsi-magnify/frontend-lib". This is made possible by the package json in the src, defining the workspace with the name "jitsi-magnify", and the sub projects.


Leonils:
The Circle CI should be updated to work with the new structure, something I'm not very confident about... I will try using the ci from richies as a reference

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
sampaccoud, line 1: you have yarn.lock and src/yarn.lock when you should have only src/frontend/yarn.lock or am I missing something?

Leonils: Yes, indeed. Run any yarn command from top level generate this empty lock. I delete it.

src/magnify/rooms/models.py
```diff
@@ -1,17 +1,12 @@
...
-
 
 class User(auth_models.AbstractUser):
     """User model which follow courses or manage backend (is_staff)"""
 
     class Meta:
         db_table = "magnify_user"
-        verbose_name = _("User")
-        verbose_name_plural = _("Users")
+        verbose_name = "Users"
```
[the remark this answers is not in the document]
Leonils: I agree. Not quite sure where translations are defined and where they should be, so I created the most minimal user model. I can try to make it work though

docs/tools.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Tests
+
+Frontend tests are written using [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/). On the frontend, unit tests are located in each component folder. They can be run using the command `yarn test` inside the `src/frontend` folder.
+
+Backend tests are located in the `tests` folder. They can be run using the command `make test-back`. It will run the backend tests inside a docker. 
+
+## Storybook
+
+To develop the components of the frontend library in isolation, you can use `yarn storybook` inside the `src/frontend` folder. It will start a storybook server that present each component in a separate page with some controls to interact with it. This works by writing a story for each component. Stories are located also in the folder of the component they present.
```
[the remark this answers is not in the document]
Leonils: Thank you for the corrections

docs/architecture.md
```diff
@@ -0,0 +1,18 @@
+# Architecture
+
+This project is divided in 4 main parts:
+
+- `src/frontend`: A frontend library, published as `@jitsi-magnify/frontend-lib` on NPM. 
+- `src/demo/frontend`: A frontend demo, published as `@jitsi-magnify/frontend-demo`: a simple react app that show a typical usecase for the components, the views, the code provided by the front library
+- `src/magnify`: A backend library, published as `magnify` on PyPI, that exposes a typical django configuration, and some apps (currently only one)
+- `src/demo/backend`: A backend demo: a django project that show how to integrate the apps in an actual project
+
+The demo projects are designed for the usecase of openfun, but can be used a a basis for your own project. The apps published on PyPI can be used independantly in your own django project.
```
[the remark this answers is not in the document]
Leonils: Yes, of course, but nevertheless, it will somehow be tailored to FUN use-case in the choices we will have to do. The libraries would be much more generic. You may use the demo out of the box with some settings, but to create a fully personalized experience, you might need to create it from the library (that will contain most of the logic)

Makefile
```diff
@@ -39,7 +39,7 @@ COMPOSE_RUN_APP      = $(COMPOSE_RUN) app-dev
 COMPOSE_RUN_CROWDIN  = $(COMPOSE_RUN) crowdin crowdin
 COMPOSE_TEST_RUN     = $(COMPOSE_RUN)
 COMPOSE_TEST_RUN_APP = $(COMPOSE_TEST_RUN) app-dev
-MANAGE               = $(COMPOSE_RUN_APP) python manage.py
+MANAGE               = $(COMPOSE_RUN_APP) python demo/backend/manage.py
```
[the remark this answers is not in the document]
Leonils: So we would mount `src/demo/backend:/app`, and where to we mount `src/magnify` in order to make `pip install -e magnify` to continue tracking magnify ?

src/frontend/src/components/TestButton/TestButton.tsx
```diff
@@ -0,0 +1,6 @@
+import React, { ReactElement } from "react";
+import classes from "./styles.module.css";
+
+export default function TestButton(): ReactElement {
```
lunika, line 4:
```suggestion
export default function TestButton() {
```

IMO there is no need to type the return of a react component.

src/frontend/src/types/modules.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.module.css" {
+  const classes: { [key: string]: string };
+  export default classes;
+}
```
lunika, line 4: What do you think about using CSS in js with a libray like https://styled-components.com/
Moreover, you plan to use Grommet and this library rely on styled component to create their component and using styles component is the best way to override a grommet component if you want to change locally its styles.

src/frontend/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
lunika, line 3: I suggest you to order these settings.

lunika (commented):
Few suggestions: 
- You can probably configure you IDE to automatically add a new line at the end of a file
- There is nothing explaining how to build the front application. Did you plan to add a target in the Makefile ? Something else in the documentation ?

Dockerfile
```diff
@@ -1,91 +1,57 @@
...
 
-# Upgrade pip to its latest release to speed up dependencies installation
-RUN python -m pip install --upgrade pip
-
-# Upgrade system packages to install security updates
-RUN apt-get update && \
-  apt-get -y upgrade && \
-  rm -rf /var/lib/apt/lists/*
+# ---- Front-end builder image ----
+FROM node:14 as front-builder
```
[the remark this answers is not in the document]
Leonils: Yes, we can.

src/frontend/src/components/TestButton/TestButton.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import TestButton from "./TestButton";
+import { render } from "@testing-library/react";
+
+describe("TestButton", () => {
+  it("renders correctly", () => {
+    const { findByText } = render(<TestButton />);
+    expect(findByText("Test")).toBeTruthy();
```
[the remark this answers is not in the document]
Leonils: Very good resource, thank you

src/frontend/tsconfig.json
lunika: Yes

src/frontend/src/components/TestButton/TestButton.tsx
lunika: The return type is inferred by typescript. 
If you absolutely want to use typing with a component, I suggest you to type the function itself using `React.FC` this way:

```
const TestButton: React.FC = () => <button className={classes.root}>Test</button>;

export default TestButton
```

lunika: And here an article you can read about typing a react component: https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript

Leonils:
@lunika, about 2 suggestions, I did it for the final line, thank you. 

About how to build the frontend, it will come soon, but I think not in this PR. The updates in this PR are required for any other work on this repository, so I think what we need quite urgently is development environment working. 

Right now, we should: 
- Run `yarn` in src folder to install dependencies
- Run `yarn build` in `src/frontend` folder to build the lib
- Run `yarn storybook` in `src/frontend` folder to develop the lib using storybook
- Run `yarn start` in `src/demo/frontend` to run the demo frontend

Leonils:
This should be ready now. Now we should have a clean development image, even if for that I had to rewrite the Dockerfile to understand it better... 

There is one interrogation however. To install the package magnify in edit mode, I copied it in the Dockerfile at `/usr/local/jitsi-magnify/magnify` and then installed it with -e. Then the docker file mount the lib in the same location, so the link is kept. However, I discussed with a developer from Illuin Tech, that was dubious about that. The pip install -e is made at built time, and the mounting is made at runtime. On my machine it works, but it might work only on linux, as MacOS and Windows add another layer of VM to run docker, that might prevent that from working. I have not access to one of these right now, so feel free to try if you can. Thank you.

src/frontend/src/components/TestButton/TestButton.tsx
```diff
@@ -1,6 +1,25 @@
...
+  color: white;
+`;
+
+const BlueButton = styled(Button)`
+  background-color: blue;
+  color: white;
+`;
+
+export interface TestButtonProps {
+  variant: "blue" | "red";
```
lunika, line 16: using an `Enum` should be better IMO

```
enum TestButtonVariant {
  BLUE = 'blue',
  RED = 'red'
}

export interface TestButtonProps {
  variant: TestButtonVariant
}
```

src/frontend/src/components/TestButton/TestButton.tsx
```diff
@@ -1,6 +1,25 @@
...
+}
+
+export default function TestButton(props: TestButtonProps) {
```
lunika, line 19:
```suggestion
export default function TestButton({ variant }: TestButtonProps) {
```

src/frontend/src/components/TestButton/TestButton.tsx
```diff
@@ -1,6 +1,25 @@
...
+  const { variant } = props;
```
lunika, line 20:
```suggestion
```

src/frontend/src/components/TestButton/TestButton.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import TestButton from "./TestButton";
+import { render, screen } from "@testing-library/react";
+
+describe("TestButton", () => {
+  it("renders correctly", () => {
+    render(<TestButton variant="red" />);
+    screen.getByText("Test Button");
```
lunika, line 8: Here you should not use this selector but `getByRole` to be sure you render a button and this button is accessible

```suggestion
    const button = screen.getByRole('button', { name: "Test Button"});
```

Then you should test it has the required variant. For this, you can install the package `@testing-library/jest-dom` that adds several asserts to jest, you can check the doc. Here you can use [toHaveStyle](https://github.com/testing-library/jest-dom#tohavestyle)

```
expect(button).toHaveStyle('color: red');
```

And to finish you should test the component with all available variants.

Leonils:
A better proposal was made by @sampaccoud with PR #8. We may close this PR.

PR #11, opened 2022-06-13
Db model

### LINKED ISSUE #5: Create a database model to store meetings, groups and users
### Abstract
The auth system should go beyond a simple token provider and store more about the users. Ideally, one could create and plan meetings. A meeting would be defined at a certain time on a certain url. It is planned for a group of guests (registered on the auth or identified by their emails) or can be open to anybody. Each user can also create and manage user groups that will be useful during meetings. 

### Technical solution
The solution would be a postgresql database created using django's model system. It implements all the features described earlier.  

### Considered alternatives
A stateless app was thought of but it clearly limits the possibilities of the app.

### Adoption strategy
The model is to be laid down on paper as to be implemented later using django's models.

### Am I going to work on it through a PR?
Most probably yes, in a very near future

DarkPitoune:
## Purpose

Resolve #5 

## Proposal

Create models according to the schema drawn in the issue #5


Makefile
```diff
@@ -216,6 +216,11 @@ migrate: ## perform database migrations
 	@$(MANAGE) migrate
 .PHONY: migrate
 
+makemigrations: ## create database migrations
```
sampaccoud, line 219: about your commit messages, it would be better to avoid merge commits. You might want to read our contribution guidelines out git messages: https://handbook.openfun.fr/git

src/magnify/apps/core/models.py
```diff
@@ -15,3 +16,98 @@ class Meta:
...
+    # Set to True if there is a meeting on that day
+    monday = models.BooleanField(default=False)
+    tuesday = models.BooleanField(default=False)
+    wednesday = models.BooleanField(default=False)
+    thursday = models.BooleanField(default=False)
+    friday = models.BooleanField(default=False)
+    saturday = models.BooleanField(default=False)
+    sunday = models.BooleanField(default=False)
+
+    start_Time = models.TimeField()
```
sampaccoud, line 37:
```suggestion
    start_time = models.TimeField()
```
But I'm not sure if this field is still necessary now that the `start` field is a datetime?

src/magnify/apps/core/models.py
```diff
@@ -15,3 +16,98 @@ class Meta:
 
     def __str__(self):
         return self.username
+
+class Meeting(models.Model):
+    """Model for one meeting or a collection of meetings defined reccursively"""
```
sampaccoud, line 21: I'm not sure how all these fields will allow scheduling single meetings or recurring meetings. I would like to be sure it's a good idea so you need to provide 2 things:
- a page in the documentation with example of each use case
- if necessary tests with examples of each use case

src/magnify/apps/core/models.py
```diff
@@ -15,3 +16,98 @@ class Meta:
...
+    name = models.CharField(max_length=100)
+
+    # Start and end are the same for a single meeting
+    start_date = models.DateTimeField()
+    end_date = models.DateTimeField()
```
sampaccoud, line 26: since it's a datetime, don't call it date. I propose to just call them:
```suggestion
    start = models.DateTimeField()
    end = models.DateTimeField()
```

src/magnify/apps/core/models.py
```diff
@@ -15,3 +16,98 @@ class Meta:
...
+    tuesday = models.BooleanField(default=False)
+    wednesday = models.BooleanField(default=False)
+    thursday = models.BooleanField(default=False)
+    friday = models.BooleanField(default=False)
+    saturday = models.BooleanField(default=False)
+    sunday = models.BooleanField(default=False)
+
+    start_Time = models.TimeField()
+    expected_duration = models.DurationField()
+    admin = models.ManyToManyField(User)
```
sampaccoud, line 39: Should be a plural and we want to avoid abbreviations:
```suggestion
    administrators = models.ManyToManyField(User)
```

src/magnify/apps/core/models.py
```diff
@@ -15,3 +16,98 @@ class Meta:
 
     def __str__(self):
         return self.username
+
+class Meeting(models.Model):
+    """Model for one meeting or a collection of meetings defined reccursively"""
+    name = models.CharField(max_length=100)
```
sampaccoud, line 22: 100 is ok for the room name but may be a bit short for the meeting name?

src/magnify/apps/core/models.py
```diff
@@ -15,3 +16,98 @@ class Meta:
...
+        verbose_name = _("Room")
+        verbose_name_plural = _("Rooms")
+    
+    def __str__(self):
+        return self.name
+
+class Group(models.Model):
+    name = models.CharField(max_length=100)
+    token_id = models.CharField(max_length=100)
+    token = models.CharField(max_length=100)
```
sampaccoud, line 66: Can you explain what these fields are for?

src/magnify/apps/core/models.py
sampaccoud: In the documentation for example:
```
We should be able to define a meeting taking place once, starting on a predefined date and time, for example June 13, 2022, and expected to last 2h. In this case, the instance would take the following values:
start = ...
end = ...
...

File: src/magnify/apps/core/models.py [hunk 8]
```diff
@@ -15,3 +16,98 @@ class Meta:
...
+    group_meeting = models.ManyToManyField(Meeting)
+    group_room = models.ManyToManyField(Room)
+    members = models.ManyToManyField(User, related_name="GroupMembers")
+
+    class Meta:
+        db_table = "magnify_group"
+        verbose_name = _("Group")
+        verbose_name_plural = _("Groups")
+
+class GroupMembers(models.Model):
```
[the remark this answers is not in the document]
DarkPitoune: It is defined using through. But the simple keyword does't allow creating a `is_administrator` attribute on the relation table.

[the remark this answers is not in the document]
sampaccoud: You set it as "related_name". I thought it's the "through" attribute. If we had a test we would know if it works :wink:

src/magnify/apps/core/models.py
```diff
@@ -15,3 +17,98 @@ class Meta:
...
     def __str__(self):
         return self.username
+
+class Meeting(models.Model):
+    """Model for one meeting or a collection of meetings defined reccursively"""
+    name = models.CharField(max_length=500)
+
+    # Start and end are the same for a single meeting
+    start = models.DateTimeField()
+    end = models.DateTimeField()
```
sampaccoud, line 27: Now that the time is a separate field, these should be of type `DateField`... I know I asked the opposite but that's when I didn't understand the model usage yet :hand_over_mouth:

src/magnify/apps/core/models.py
```diff
@@ -15,3 +17,98 @@ class Meta:
...
+    
+    # The following fields are to be changed when we learn more about what is configurable
+    is_open = models.BooleanField(default=False)
+    has_microphone_by_default = models.BooleanField(default=False)
+    has_camera_by_default = models.BooleanField(default=False)
+
+    class Meta:
+        db_table = "magnify_jitsi_configuration"
+        verbose_name = _("Jitsi Configuration")
+        verbose_name_plural = _("Jitsi Configurations")
```
sampaccoud, line 114: missing line break

src/magnify/apps/core/models.py
```diff
@@ -15,3 +17,98 @@ class Meta:
...
+    class Meta:
+        db_table = "magnify_room"
+        verbose_name = _("Room")
+        verbose_name_plural = _("Rooms")
+    
+    def __str__(self):
+        return self.name
+
+class Group(models.Model):
+    name = models.CharField(max_length=100)
```
sampaccoud, line 65: Not critiacal but you're missing a few docstrings

sampaccoud (approved):
A few neatpicks but otherwise GTM :tada:

PR #12, opened 2022-06-14
Group list components

Leonils:
## Purpose

Add components to present a selectable list of groups. Actions are not implemented yet. The components are responsive. 

## Proposal

3 components added:
- A single group row: `GroupRow`
- A header to select/unselect several rows `GroupsHeader`
- The list itself, including the 2 previous components `GroupsList`

Screenshot from 2022-06-14 16-27-45
Screenshot from 2022-06-14 16-27-55


src/frontend/src/components/groups/GroupRow/GroupRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { useMediaQuery } from 'react-responsive';
+
+export interface Group {
+  id: string;
+  name: string;
+  members: {
+    id: string;
+    name: string;
+    avatar: string;
+  }[];
```
lunika, line 13: `members` can be extracted as a single interface ? Then you use it in you `Group` interface.

```suggestion
  members: Array<{
    id: string;
    name: string;
    avatar: string;
  }>;
```

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+   */
+  onToogle: (selected: boolean) => void;
+  /**
+   * Is the group currently selected
+   */
+  selected: boolean;
+}
+
+export default function GroupRow(props: GroupRowProps) {
+  const { group, onToogle, selected } = props;
```
lunika, line 32:
```suggestion
```

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const ref = useRef<HTMLInputElement>(null);
```
lunika, line 33: This is not true. The ref contains an `HTMLInputElement` or `null`. 
I suggest you to create some type utils and declare these two you will use a lot:

```
export type Maybe<T> = T | undefined;

export type Nullable<T> = T | null;
```

And then here 

```suggestion
  const ref = useRef<Nullable<HTMLInputElement>>(null);
```

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+export interface GroupRowProps {
+  /**
+   * The group to display
+   */
+  group: Group;
+  /**
+   * Callback when the group is selected or deselected
+   */
+  onToogle: (selected: boolean) => void;
```
lunika, line 24:
```suggestion
  onToggle: (selected: boolean) => void;
```

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+   * Is the group currently selected
+   */
+  selected: boolean;
+}
+
+export default function GroupRow(props: GroupRowProps) {
+  const { group, onToogle, selected } = props;
+  const ref = useRef<HTMLInputElement>(null);
+
+  const handleToogle = (event: ChangeEvent<HTMLInputElement>) => {
```
lunika, line 35:
```suggestion
  const handleToggle = (event: ChangeEvent<HTMLInputElement>) => {
```

src/frontend/src/components/groups/GroupRow/GroupRow.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+      render(
+        <GroupRow
+          group={{ id: '1', name: 'Group 1', members: mockedMembers }}
+          selected={initialChecked}
+          onToogle={mockedToogle}
+        />,
+      );
+
+      fireEvent.click(screen.getByTitle('Select Group'));
```
lunika, line 75: You should use [useEvent](https://testing-library.com/docs/user-event/intro) instead of `fireEvent`

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  // Get a number between 0 and 3 representing the size of the screen
+  // 0 = big, 1 = large, 2 = medium, 3 = small, 4 = xsmall
+  const isBigScreen = useMediaQuery({ query: '(min-width: 1824px)' });
+  const isLargeScreen = useMediaQuery({ query: '(min-width: 1224px)' });
+  const isMediumScreen = useMediaQuery({ query: '(min-width: 824px)' });
+  const isSmallScreen = useMediaQuery({ query: '(min-width: 424px)' });
+  const matchingScreenSizes = [isBigScreen, isLargeScreen, isMediumScreen, isSmallScreen, true];
+  const screenSize = matchingScreenSizes.findIndex((size) => size);
+  const numberOfDisplayedMembers = 9 - 2 * screenSize;
```
lunika, line 47: I don't know if all of this is not overkill. First grommet contains a responsive context you can use. You should also use its provider to make you app responsive.

src/frontend/src/components/groups/GroupRow/GroupRow.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+        screen.getByTitle(member.name);
+      });
+
+      // The title of the group and the number of participants should be displayed
+      screen.getByText('Group 1');
+      screen.getByText(numberOfMembers.toString());
+
+      // If all images are not displayed, we should have an indicator with remaining names
+      if (allDisplayed) {
+        expect(screen.queryByTestId('more-members')).toBeNull();
```
lunika, line 48:
```suggestion
        expect(screen.queryByTestId('more-members')).not.toBeInTheDocument();
```

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
Leonils: Indeed. In this case I need to remove the ref completely as it is useless, but I keep the ideal of such generics. Thank you

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
Leonils: I'm not used to destructure props directly, but we can commit to that indeed.

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
Leonils: In your opinion, where such interfaces should go ? In the `/src/types`, (very global), in a `/src/components/groups/types.ts` (root of where we need it, but it adds files randomly), or in one of the components that use it, as it is right now...

src/frontend/src/components/groups/GroupRow/GroupRow.tsx
Leonils: Yes, maybe. I'll look at what grommet provide... If it exposes the theme-defined breakpoints it would be much better indeed !

src/frontend/src/components/groups/GroupRow/GroupRow.test.tsx
Leonils: Indeed, the behaviour is much better. For instance, the focus is lost when clicked in something else. Thank you.

PR #14, opened 2022-06-16
Init database

DarkPitoune:
This PR replaces #11 and solves #5


Makefile
```diff
@@ -148,6 +148,13 @@ migrate:  ## run django migrations for the magnify project.
 	@$(MANAGE) migrate
 .PHONY: migrate
 
+makemigrations: ## run django makemigrations for the magnify project.
```
sampaccoud, line 151: in your commit message : s/modsel/models

src/backend/magnify/core/models.py
```diff
@@ -15,3 +17,98 @@ class Meta:
...
+
+    class Meta:
+        db_table = "magnify_label"
+        verbose_name = _("Label")
+        verbose_name_plural = _("Labels")
+
+class JitsiConfiguration(models.Model):
+    """Model for the Jitsi configuration of a room or a meeting"""
+    meeting = models.ForeignKey(Meeting, on_delete=models.CASCADE)
+    room = models.ForeignKey(Room, on_delete=models.CASCADE)
```
sampaccoud, line 104: My understanding is that it will be either linked to a meeting or linked to a room but not both?
in this case you need to make these fields nullable:
```suggestion
    meeting = models.ForeignKey(Meeting, on_delete=models.CASCADE, null=True, blank=True)
    room = models.ForeignKey(Room, on_delete=models.CASCADE, null=True, blank=True)
```

PR #15, opened 2022-06-16
✨(models) add models for groups, user, meetings, labels

### LINKED ISSUE #5: Create a database model to store meetings, groups and users

DarkPitoune:
Resolve #5 
Create new models for meetings, labels, groups, configuration of the meetings.


sampaccoud (approved):
Don't forget to add a line in the changed log:

## Added

- Add models to handle rooms, meetings and groups

PR #16, opened 2022-06-16
Group components

Leonils:
## Purpose

Add components to present a selectable list of groups. Actions are not implemented yet. The components are responsive.

## Proposal

3 components added:

- A single group row: `GroupRow`
- A header to select/unselect several rows `GroupsHeader`
- The list itself, including the 2 previous components `GroupsList`

Remark : this is the continuation of the PR #12, for which the branch was too much different from main after the refactoring from #8


sampaccoud:
I have 2 general comments:
- now that we added ReactIntl, shouldn't we translate all our strings?
- using factories in your tests would be more powerful than mocks? See here for example: https://github.com/openfun/richie/blob/master/src/frontend/js/utils/test/factories.ts

Leonils:
For the factories, it may be useful, but maybe it can come along with the mocking of the requests to the api ( see #19 ). Another concern is that the reference lib to do that in js has been violently discontinued by it's creator (faker), and the example you mentioned has this one as a dependency. Should we use it anyway ? Has anyone an alternative ?

AlexLaval:
> the reference lib to do that in js has been violently discontinued by it's creator (faker)

It _seems_ that Faker has found new maintainers, but I don't know if this is the version used in factories.
Github : https://github.com/faker-js/faker
Announcement : https://fakerjs.dev/about/announcements/2022-01-14.html

src/frontend/magnify/package.json
```diff
@@ -32,6 +32,7 @@
     "@storybook/testing-library": "0.0.11",
     "@testing-library/jest-dom": "5.16.4",
     "@testing-library/react": "13.3.0",
+    "@testing-library/user-event": "^14.2.0",
```
lunika, line 35:
```suggestion
    "@testing-library/user-event": "14.2.0",
```

src/frontend/magnify/src/components/groups/GroupRow/GroupRow.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+          />
+        </ResponsiveContext.Provider>,
+      );
+
+      // Check if the number of images rendered is correct
+      const avatars = screen.queryAllByTestId('group-row-member-image');
+      expect(avatars.length).toBe(expectedNumberOfMembersRendered);
+
+      // Check if images have the right title
+      mockedMembers.slice(0, expectedNumberOfMembersRendered).forEach((member, index) => {
```
lunika, line 42:
```suggestion
      mockedMembers.slice(0, expectedNumberOfMembersRendered).forEach((member) => {
```

src/frontend/magnify/src/components/groups/GroupRow/mocks.ts
```diff
@@ -0,0 +1,17 @@
+const names = ['John', 'Jane', 'Bob', 'Mary', 'Jack', 'Jill', 'Adam', 'Julie', 'Bobby'];
```
lunika, line 1: As suggested in a comment you can use fakerjs for this (the new one) https://github.com/faker-js/faker

src/frontend/magnify/src/types/maybe.type.ts
```diff
@@ -0,0 +1 @@
+export type Maybe<T> = T | undefined;
```
lunika, line 1: You can group together `Maybe` and `Nullable` in `misc` file, you will probably have other type like these

lunika: and IMO you can remove `type` in the file name, you are already in a type directory

src/frontend/magnify/src/types/group.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { Member } from './member.interface';
```
lunika, line 1: Adding `interface` in the file name is useless IMHO. You are in a `types` directory, it's already explicit.

lunika (approved):

PR #27, opened 2022-06-22
Create components for user page

Leonils:
## Purpose

This add the components and main blocks from the profile page. The "linked accounts" was skipped as it is still under discussion on whether or not we should include GAFAM providers

## Proposal

This add following components

- Forms
  - AvatarForm
  - IdentityForm (name, username, email)
  - PasswordUpdateForm
- Blocks: a block is a section of the page, grouping some forms/controls
  - IdentityBlock (grouping avatar and identity form)
  - PasswordUpdateBlock 
  - DeleteAccountBlock

But also some genericity, with a custom TextField component, and a generic hook to handle the form state

Finaly, here are some screenshots of the 3 blocks
 
Screenshot from 2022-06-22 12-01-28

### DIFF
--- a/src/frontend/magnify/src/components/design-system/TextField/TextField.test.tsx
+++ b/src/frontend/magnify/src/components/design-system/TextField/TextField.test.tsx
@@ -20,7 +20,7 @@ describe('TextField', () => {
 
     render(<TestElement />);
 
-    const input = screen.getByLabelText('My input') as HTMLInputElement;
+    const input = screen.getByRole<HTMLInputElement>('textbox', { name: 'My input' });
 
     await user.type(input, ' World');
     expect(input.value).toBe('Hello World');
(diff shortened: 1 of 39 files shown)


src/frontend/magnify/src/components/design-system/TextField/TextField.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+          name="my-input"
+          onChange={(e) => setValue(e.target.value)}
+          value={value}
+        />
+      );
+    };
+
+    render(<TestElement />);
+
+    const input = screen.getByLabelText('My input') as HTMLInputElement;
```
lunika, line 23: Perhaps if you get it by role you don't have to cast the type and I think it's better to do it this way.

src/frontend/magnify/src/components/profile/IdentityForm/IdentityForm.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <IdentityForm name="John Doe" username="johndoe3" email="john.doe@test.fr" />
+      </IntlProvider>,
+    );
+
+    const nameInput = screen.getByLabelText('Name*');
+    const usernameInput = screen.getByLabelText('Username*');
+    const emailInput = screen.getByLabelText('Email*');
+
+    // Save button is initially disabled (no modification)
+    const saveButton = screen.getByText('Save');
```
lunika, line 22: Prefer `getByRole` here

PR #30, opened 2022-06-22
Create routes for user creation and login 

DarkPitoune:
[This PR is a bit urgent as it will unlock development on the front end]
## Purpose

To create a user and to login as a user

## Proposal
There is one view for each route. The auth takes as much of the native features of django as possible


sandbox/settings.py
```diff
@@ -95,6 +95,24 @@ class Base(DRFMixin, MagnifyCoreConfigurationMixin, Configuration):
...
     AUTH_USER_MODEL = "core.User"
 
+    AUTH_PASSWORD_VALIDATORS = [
+        {
+            "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",
+        },
+        {
+            "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
+            "OPTIONS": {
+                "min_length": 8,
```
lunika, line 105: Add the ability to override this value with an environment variable using `values.Value`

src/magnify/apps/core/models.py
```diff
@@ -8,7 +8,48 @@
...
+        "^[a-zA-Z0-9_-]{3,16}$",
+        message="""Username must be between 3 and 16 characters long
+            and contain only letters, numbers, underscores and hyphens.""",
+    )
+
+    username = models.CharField(
+        _("username"),
+        max_length=150,
+        unique=True,
+        validators=[username_validator],
```
lunika, line 23: I'm not fan of having validation on the model. In an API I prefer having them on the serializer unless you plan to use this model in a management command.

src/magnify/apps/core/models.py
```diff
@@ -8,7 +8,48 @@
...
+            RegexValidator(
+                regex=r"^[a-zA-Z '-]+$",
+                message=_("Only letters, spaces, apostrophe and hyphen are allowed"),
+            )
+        ],
+    )
+    email = models.EmailField(
+        max_length=255,
+        unique=True,
+        validators=[
```
lunika, line 43: IMO this is useless and already managed by DRF and django https://www.django-rest-framework.org/api-guide/fields/#emailfield

src/magnify/apps/core/serializers/users.py
```diff
@@ -0,0 +1,42 @@
+"""Serializers for users views"""
+
+from rest_framework import serializers
+
+
+class UserCreateSerializer(serializers.Serializer):
```
lunika, line 6: Why not using a ModelSerializer mapping it with the user model ? Doing this you have only one serializer https://www.django-rest-framework.org/api-guide/serializers/#modelserializer

src/magnify/apps/core/views/user.py
```diff
@@ -0,0 +1,87 @@
...
+
+from magnify.apps.core.models import User
+from magnify.apps.core.serializers import (
+    UserCreateResponseErrorSerializer,
+    UserCreateResponseSuccessSerializer,
+    UserCreateSerializer,
+)
+
+
+class UserView(APIView):
```
lunika, line 19: Consider using a ModelViewset to not implement everything https://www.django-rest-framework.org/api-guide/viewsets/#modelviewset

lunika:
Lot of code in this PR can be removed using DRF views and serializer mapped on the model.

src/magnify/apps/core/models.py
sampaccoud: why are you against model validation for this case? Shouldn't it even be a database constraint if we really want to be sure our users are properly named?

lunika: Most of time validators on the database avoid future modification. Once modified, part of your data will become invalid and you will not be able to modify the all model without doing a sql migration.

sampaccoud: I would consider this a feature but we can discuss irl.

DarkPitoune:
@sampaccoud I added jwt auth on some routes so the test now fail. They are not required but it would be nice if it was patched before merge

sandbox/settings.py
```diff
@@ -208,6 +235,7 @@ class Base(DRFMixin, MagnifyCoreConfigurationMixin, Configuration):
         "dockerflow.django",
         "parler",
         "rest_framework",
+        "rest_framework.authtoken",
```
sampaccoud, line 238: no more needed

src/magnify/apps/core/urls.py
```diff
@@ -13,15 +16,25 @@
         title="Magnify API",
         default_version="v1",
         description="""This is the schema for the Magnify API.
-            This app is used in the sandbox folder of the project.""",
+
+            All routes are protected by the JWT authentication except
+            for "login", "refresh" and "create user".""",
     ),
     public=True,
+    url=os.environ.get("BACK_URL"),
```
sampaccoud, line 24: you should go through a Django setting (see existing settings with value.Value())

src/magnify/apps/core/views/room_token_view.py
```diff
@@ -53,8 +54,11 @@ class RoomTokenView(APIView):
     """View for token fetching based on a guest user and a specific room
     This redirects to jitsi with the token"""
 
+    permission_classes = (IsAuthenticated,)
```
sampaccoud, line 57: This is why your tests are failing. You should not force authentication, you should test in the view that if the user is connected and deduct from this what s.he can do or not.

src/magnify/apps/core/models.py
DarkPitoune: I really don't know how to add validators in the serializer. If the feature is seen as necessary, some help would be welcome ;)

src/magnify/apps/core/serializers/users.py
DarkPitoune: Tks a lot for this review, I'm working on it
I still have some issues with the response from the API. In order to have good documentation, I'm creating a serializer for each possible response, which is very heavy. Do you know if there is a better way to handle this or should I just not care ?

sandbox/settings.py
DarkPitoune: A value does not work for it triggers an error when django tries to compare an integer and an intergerValue. It needs an integer

DarkPitoune:
All reviews were implemented. Everything is tested by hand but avatar. I'm writing tests to ensure everything but I'm ready for reviews on the code :)

DarkPitoune:
PR is fully completed and tested. I don't plan on writing the avatar endpoints and tests on this PR
Open for reviews

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Add models to handle rooms, meetings, labels and groups
 - Add a swagger view to help with front end development
+- Add a route for user creation and login
```
kernicPanel, line 13:
```suggestion
- Add routes for user creation, modification and login
```

kernicPanel (approved):

DarkPitoune:
PR is unusable as is, another one is to be opened soon

Issue #28, opened 2022-06-22
Dev scripts to help development


AlexLaval:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
When creating frontend components for example, a huge part of the code is very similar from one component to another. Finding a way to automate folder and files creation could be useful.

**Describe the solution you'd like**
Some scripts, for example in `src/frontend/scripts/`, that can be executed to add components.

**Describe alternatives you've considered**
Keeping on doing it by hand, but it's not very convenient.

**Discovery, Documentation, Adoption, Migration Strategy**
You should be able, in `src/frontend` folder, to execute a script like `yarn add-icon` that could give you all necessary files in the right place in the `magnify` folder, and updates export references and stories.

**Do you want to work on it through a Pull Request?**
Yes.

sampaccoud:
:medal_sports: @AlexLaval lazy devs are the best! :smile:

PR #31, opened 2022-06-22
Dev scripts to build icons and components

AlexLaval:
## Purpose

As precised on Issue #28, creating all necessary folders and files by hand for frontend components is tedious.
This PR aims to change that by adding scripts to execute on the dev side.

## Proposal

You have 2 scripts available, and room for adding more:

- `yarn scripts:add-icon <IconNameSVG>` to add a SVG Icon template at the right place. You just have to paste a SVG in one file, by following comments.
- `yarn scripts:add-component <Path/To/Your/ComponentFolder>` to add a component, all its necessary files (including storybook stories and tests), and to update imports in subfolders' `index.ts`.


src/frontend/package.json
```diff
@@ -26,7 +26,9 @@
     "format-magnify:write": "cd magnify && yarn format:write",
     "format:check": "yarn format-magnify:check && yarn format-demo:check",
     "format:write": "yarn format-magnify:write && yarn format-demo:write",
-    "verify-all": "yarn format:check & yarn lint & yarn build & yarn test"
+    "verify-all": "yarn format:check & yarn lint & yarn build & yarn test",
```
kernicPanel, line 29: Are you sure you want parallel processing here ?
Maybe you want to run commands only if the previous one succeed (personally I prefer this):
```suggestion
    "verify-all": "yarn format:check && yarn lint && yarn build && yarn test",
```
Or you want to run them all anyway
```suggestion
    "verify-all": "yarn format:check; yarn lint; yarn build; yarn test",
```

kernicPanel (commented):
Very nice script !
Some comments after a quick test :
You should output an explicit usage when arguments are missing
For the component generator, I think you should enforce capitalization of the argument

I'll continue this review more deeper.

src/frontend/scripts/src/tools/appendToFile.ts
```diff
@@ -0,0 +1,39 @@
...
+  after?: boolean
+) {
+  try {
+    if (!stringToFind) {
+      fs.appendFileSync(filePath, line);
+      console.log(`Appended line in ${path.basename(filePath)}`);
+      return;
+    }
+
+    var data = fs.readFileSync(filePath).toString().split("\n");
```
kernicPanel, line 18:
```suggestion
    const data = fs.readFileSync(filePath).toString().split("\n");
```
There are others `var` in this file and others, please use `const` or `let` instead.

src/frontend/scripts/src/generateIcon.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// ICONNAME.TSX
+writeFile(path.join(fullPath, `${iconName}.tsx`), iconTemplate(iconName));
+
+// EXPORT IN ICONS' INDEX.TS
+appendToFile(path.join(fullPath, "../index.ts"), exportTemplate(iconName));
+
+// ADD ICON IN STORY
+appendToFile(
+  path.join(fullPath, "../SVGIcon.stories.tsx"),
```
kernicPanel, line 33: This file seems to be missing, did I forgot something to do?

kernicPanel (commented):
Very cool scripts :clap: 
Please consider my previous comments.

src/frontend/scripts/src/generateIcon.ts
kernicPanel: Ok, I see, thanks for your clarification.
Maybe you could edit this PR to target [sidebar-frontend-components](https://github.com/openfun/jitsi-magnify/tree/sidebar-frontend-components) instead of [main](https://github.com/openfun/jitsi-magnify) then ?

src/frontend/package.json
AlexLaval: Since this is already merged, what do you think we should do @kernicPanel ? Commit this on this PR ? A new Issue just for this line (😅) ? Or maybe a `--fixup` on `main` ?

kernicPanel: As this command is useful to check the generated components, I think we can fix it directly  in this PR.
Otherwise, a separate PR would be needed.

Doing a fixup on main will force us to make a choice:
- leaving this fixup commit in the repository history, which would be extremely ugly.
- squash the fixup on the related commit, so rewrite the main branch history, which could lead to problems with your coworkers.
So none of any choices are good, you should never do a fixup on a main branch.
They are only allowed on commit that can be squashed.
Let me know if something is not clear about this point. I think it's really important to get it.

AlexLaval: That's perfectly clear and makes a lot of sense, thanks !

PR #35, opened 2022-06-27
✨(frontlib) create a row to display a single meeting

Leonils:
## Purpose

Create a row to display a meeting

## Proposal

Create a component to display all needed information, and some logic to handle "joining a meeting", and to extract to the pattern of meetings the actual next meeting

Screenshot from 2022-06-27 16-12-42
Screenshot from 2022-06-27 16-12-52
Screenshot from 2022-06-27 16-12-59


kernicPanel:
Not really a review, but more a general comment: I think you should split this commit into several ones.

src/frontend/magnify/jest-setup.ts
```diff
@@ -1,3 +1,7 @@
 import '@testing-library/jest-dom';
 
 Object.defineProperty(window, 'scrollTo', { value: jest.fn, writable: true });
+
+module.exports = async () => {
+    process.env.TZ = 'UTC';
```
lunika, line 6: Good idea. You can also define the environment variable `TZ` in the `test` script defined in your package.json

```
"test": "TZ=UTC jest"
```

src/frontend/magnify/src/components/meetings/MeetingRow/computeEnd.test.tsx
```diff
@@ -0,0 +1,12 @@
+import { computeEnd } from './computeEnd';
```
lunika, line 1: You can rename the extension file in `ts`

src/frontend/magnify/src/components/meetings/MeetingRow/getNextMeeting.test.tsx
```diff
@@ -0,0 +1,159 @@
+import { Hold } from '../../../factories/meeting';
```
lunika, line 1: Same here, you can rename the extension file in `ts`

src/frontend/magnify/jest-setup.ts
Leonils: I cannot mock dates without that (or headaches). I think it is not only convenient, but required for some of my tests (I mock some dates, so at this time, I choose a TZ, so it must be specified everywhere). Do you think it is safe enough to put it in the scripts ? It can easily be forgotten when writing scripts for other environments, does it ?

lunika: while you are using `yarn test` you will always use `TZ=UTC`. So no there is no risk but you can keep it like this.

kernicPanel (approved):
I'm not saying you should use it right now, and maybe you already discarded it, but I think Luxon could have been helpful for many logic of this PR: https://moment.github.io/luxon/api-docs/index.html
(especially for edge cases like calculating an end time for a meeting of 3 hours beginning at 11PM ;) )

Leonils:
About Luxon, I'm always asking myself about this kind of library. They are really helpful but usually only on a very specific part of an application, and we use a very small part of the library itself... It may save one hour, but not convinced yet...

PR #36, opened 2022-06-27
Add tests for models and optimize modelisation

sampaccoud:
## Purpose

We want to add a full test suite on models.

## Proposal

- I started by adding factories on each model
- While doing so, I had the chance to optimize a few things on the modelization:
    - remove "administrators" field on the Group model as it is already covered by the "members" field + the "is_administstrator" property on the membership,
    - The Label model had ManyToMany fields with Meetings, Rooms and Groups but it is more natural to do it the other way around because what we want is to add labels to meetings, rooms and groups,
    - add explicit ordering for all models
    - add database constraint to ensure the start date is always before the end date
    - add validation on save for all models

After doing all these optimizations, I added a full test suite to battle test factories and models.


src/magnify/apps/core/models.py
```diff
@@ -4,29 +4,72 @@
...
 from django.db import models
+from django.db.models import F, Q
 from django.utils.translation import gettext_lazy as _
 
 
-class User(auth_models.AbstractUser):
+class ValidateModelMixin:
+    """Make `save` call `full_clean`.
+
+    Should be the left most mixin in your moodels.
```
DarkPitoune, line 14:
```suggestion
    Should be the left most mixin in your models.
```

tests/apps/core/test_core_factories_labels.py
```diff
@@ -0,0 +1,18 @@
...
+class LabelsFactoriesTestCase(TestCase):
+    """
+    Unit test suite to validate the behavior of the Label factory
+    """
+
+    def test_factories_labels_color(self):
+        """The color field should be populated with a valid color."""
+        label = LabelFactory()
+        self.assertTrue(label.color.startswith("#"))
+        self.assertTrue(len(label.color), 7)
```
DarkPitoune, line 18: I don't know how to write this in an assert but the regex accepts colors like #111

DarkPitoune (changes requested):

tests/apps/core/test_core_factories_labels.py
sampaccoud: Yes but our factory only generates colors with 6 digits if I judge by this: https://github.com/joke2k/faker/blob/master/faker/providers/color/fr_FR/__init__.py

Any way we are not trying to test what the model accepts but that the factory generates valid random colors, so it should be good enough?

src/magnify/apps/core/models.py
sampaccoud: Good catch. Fixed it!

DarkPitoune (approved):

PR #38, opened 2022-06-30
Add api endpoint for rooms

sampaccoud:
## Purpose

We want to add an API endpoint to manage rooms.

## Proposal

- make many to many fields optional on all models
- make room slug unique and automatically populated
- add Simple JWT DRF for frontend authentication
- add API endpoint for rooms
- add complete test suite on the new rooms API endpoint


src/magnify/apps/core/permissions.py
```diff
@@ -0,0 +1,24 @@
...
+    Allow a request to proceed only if the user is listed as admin to the targeted object.
+    """
+
+    def has_permission(self, request, view):
+        """
+        Allow the request.
+        Only if the playlist exists and the current logged in user is one
+        of its administrator.
+        """
+        if view.action in ["create", "list", "retrieve"]:
```
lunika, line 18: Why filtering the action here ? Why not using this permission in your view only for the action you want ? I don't know if it's the permission responsibility to do this.

tests/apps/core/test_api_rooms.py
```diff
@@ -0,0 +1,232 @@
...
+
+    @mock.patch(
+        "magnify.apps.core.serializers.rooms.generate_token", return_value="the token"
+    )
+    def test_api_rooms_retrieve_administrator(self, mock_token):
+        """Administrators of a room should get an admin token."""
+        admin = UserFactory()
+        room = RoomFactory(administrators=[admin])
+        user = UserFactory()
+        jwt_token = AccessToken.for_user(user)
```
lunika, line 87: Should not be the admin here ?

tests/apps/core/test_api_rooms.py
```diff
@@ -0,0 +1,232 @@
...
+            response.json(),
+            {
+                "administrators": [admin.id],
+                "id": room.id,
+                "name": room.name,
+                "slug": room.slug,
+                "token": "the token",
+            },
+        )
+        mock_token.assert_called_once_with(user, room.slug)
```
lunika, line 103: How can you assert this is an admin token ?

tests/apps/core/test_api_rooms.py
sampaccoud: Well spotted :clap:  This test is not needed anymore since I finally chose to mock the token generation.

src/magnify/apps/core/permissions.py
sampaccoud: I would say these checks are indeed related to this permission which applies to all objects with related "administrators".

sampaccoud: Ouch the checks were wrong and I found another method to check for object permissions. It's all in a fixup.

lunika (approved):

DarkPitoune (approved):
Looks good as far as I know django ><

PR #42, opened 2022-07-01
Create login flow and speed up to a first demo

Leonils:
## Context

We are running a bit out of time, and there is still a lot to do... This PR intends to create the controller functions and to connect the whole signin flow of the user in order to integrate as many components as possible in a first demo

## Proposal

- Login, signup forms
- auth views, guards, hooks
- auth controller functions, including some advanced mocks inside the log controller
- rework of existing components to integrate them into the demo to which we can login, navigate, see groups, profile view (to be demonstrated on monday)

Hope this will work for you.


kernicPanel (approved):
As time is running out, I quickly read all this "small" PR ;)
Everything looks good to me :clap:  :clap: :clap: 

Don't forget to configure nginx to allow page reloading, as is return a 404 for now.

Leonils:
> Don't forget to configure nginx to allow page reloading, as is return a 404 for now.

Not sure what it means. @sampaccoud, do you want to look at it ?

PR #44, opened 2022-07-04
Add fallback for avatar

Leonils:
If a user has no avatar, we should display something.
Here is an attempt to display his first letters

Screenshot from 2022-07-04 15-42-50

### DIFF
--- a/src/frontend/magnify/src/components/design-system/SquareAvatar/SquareAvatar.tsx
+++ b/src/frontend/magnify/src/components/design-system/SquareAvatar/SquareAvatar.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { More } from 'grommet-icons';
-import { Avatar } from 'grommet';
+import { Avatar as GrommetAvatar, AvatarExtendedProps } from 'grommet';
 
 export interface SquareAvatarProps {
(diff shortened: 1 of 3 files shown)


src/frontend/magnify/src/components/design-system/SquareAvatar/SquareAvatar.tsx
```diff
@@ -10,9 +10,28 @@ export interface SquareAvatarProps {
...
+    <Avatar
+      round="xsmall"
+      src={src !== '' ? src : undefined}
+      size="40px"
+      title={title}
+      data-testid={dataTestId}
+      background={background}
+    >
       {more && <More color="brand" />}
+      {shouldFallback && initials}
```
kernicPanel, line 34: Just a note : those two lines lets us think that both `more` and  `shouldFallback` could be true at the same time, rendering both More and initials.
I think a good old `if` at the top of this component could be used to define a `content` variable which would be put as a child of the `Avatar` component.
Something like 
```
let content;
if(more) {
    content = (<More color="brand" />)
} else {
    content = […]
}

return (
    <Avatar […]>
        {content}
    </Avatar>
)

PR #46, opened 2022-07-04
Refactor the user model and add a user api endpoint

sampaccoud:
## Purpose

We want to start connecting users from the frontend to the backend.

## Proposal

- Refactor the user model to fit what is expected from the frontend
- Add an API endpoint for the user
- Add a full test suite for the new user API endpoint


src/magnify/apps/core/models.py
```diff
@@ -22,15 +24,66 @@ def save(self, *args, **kwargs):
...
+        _("staff status"),
+        default=False,
+        help_text=_("Designates whether the user can log into this admin site."),
+    )
+    is_active = models.BooleanField(
+        _("active"),
+        default=True,
+        help_text=_(
+            "Designates whether this user should be treated as active. "
+            "Unselect this instead of deleting accounts."
```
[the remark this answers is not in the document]
sampaccoud: This is Django default way of doing things. The thing is that sometimes it is complicated to delete a user with many related objects. In our case I think you're right, we want to allow deleting users.

src/magnify/apps/core/api.py
```diff
@@ -31,6 +39,86 @@ def exception_handler(exc, context):
     return drf_exception_handler(exc, context)
 
 
+class UserViewSet(
+    mixins.CreateModelMixin,
+    mixins.DestroyModelMixin,
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
```
lunika, line 47:
```suggestion
     viewsets.ModelViewSet
```

lunika (approved):

src/magnify/apps/core/api.py
sampaccoud: Can't be done anymore :wink:

sampaccoud:
@lunika on the list endpoint I required to pass a search query and throttled the failed search attempts aggressively so one can not look for a username or email by repeated calls to this API.

Issue #39, opened 2022-06-30
Room configuration components


asynkr:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
As an admin of a room, I must be able to change somme settings.

**Describe the solution you'd like**
Some key setting components like switches, buttons, presented in a view. Requests must be made to the API when changing settings.

**Discovery, Documentation, Adoption, Migration Strategy**
Here is a [design](https://share.goabstract.com/cb3c4b94-209e-4e33-920b-fccd88615034?collectionLayerId=1bad65e3-5bc8-4431-872d-a6b7708b319b&mode=design&present=true).

**Do you want to work on it through a Pull Request?**
Yup

PR #49, opened 2022-07-06
Room configuration components

asynkr:
## Purpose

Implement components to manage room settings as discussed on issue #39.

## Proposal

Provide several components and sections.

- Toggle components
  - A toggle component
  - A toggle component specialized for rooms
  - A Toggle block component

One difficulty is that we do not know which settings will be done on the jitsi side, and which settings we will implement in the back.
For now I added 7 setings that come from the design.

### DIFF
--- a/src/frontend/demo/src/views/roomSettings.tsx
+++ b/src/frontend/demo/src/views/roomSettings.tsx
@@ -1,0 +1,24 @@
+import { LayoutWithSidebar, RoomConfig, useTranslations } from '@jitsi-magnify/core';
+import { defineMessages } from 'react-intl';
+import { Navigate, useParams } from 'react-router-dom';
+
+export const messages = defineMessages({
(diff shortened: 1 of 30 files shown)


src/frontend/demo/src/views/roomSettings.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export default function RoomSettingsView() {
+  const intl = useTranslations();
+  const { roomName } = useParams();
+
+  if (!roomName) return <Navigate to="/room-not-found" replace />;
+
+  return (
+    <LayoutWithSidebar title={intl.formatMessage(messages.roomSettingsTitle)}>
+      <RoomConfig roomName={roomName || 'def'} />
```
kernicPanel, line 21: With the if at line 17, no default value could be used.
```suggestion
      <RoomConfig roomName={roomName} />
```

PR #53, opened 2022-07-06
Reduce size of cover image

Leonils:
## Purpose

Image has embeded png inside it. We worked with JohCoda to make
it smaller by removing embeded images but including other particles
on the border. This cut the svg size by 2


sampaccoud (approved):
:clap:

PR #54, opened 2022-07-06
Create a form to add groups to room

Leonils:
This form display the user's groups and the user
can add them to a select list. Then when he submits
the groups ids, they are added to the room

Screenshot from 2022-07-06 16-05-01

### DIFF
--- a/src/frontend/magnify/src/components/rooms/AddGroupToRoomDialog/AddGroupToRoomDialog.tsx
+++ b/src/frontend/magnify/src/components/rooms/AddGroupToRoomDialog/AddGroupToRoomDialog.tsx
@@ -1,0 +1,148 @@
+import { defineMessages, useIntl } from 'react-intl';
+import React, { useState } from 'react';
+import { Box, Button, Heading, Layer, Select, Text } from 'grommet';
+import { FormClose } from 'grommet-icons';
+import { useMutation, useQuery } from 'react-query';
+
+import { useController } from '../../../controller';
+import { Group } from '../../../types/group';
+import { LoadingButton, WaitingRow } from '../../design-system';
+
+const messages = defineMessages({
+  AddGroupToRoomDialogTitle: {
+    id: 'components.rooms.AddGroupToRoomDialog.AddGroupToRoomDialogTitle',
+    defaultMessage: 'Add groups to room',
+    description: 'Title of the add group to room dialog',
+  },
+  selectGroupsToAdd: {
+    id: 'components.rooms.AddGroupToRoomDialog.selectGroupsToAdd',
+    defaultMessage: 'Select groups to add',
+    description: 'Label for the select groups to add',
+  },
+  addGroupsLabel: {
+    id: 'components.rooms.AddGroupToRoomDialog.addGroupsLabel',
+    defaultMessage:
+      'Add {numberSelected} {numberSelected, plural, =0 {group} one {group} other {groups}}',
+    description: 'Label for the add groups button',
+  },
+  cancelLabel: {
+    id: 'components.rooms.AddGroupToRoomDialog.cancelLabel',
+    defaultMessage: 'Cancel',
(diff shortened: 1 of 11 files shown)


src/frontend/magnify/src/components/rooms/AddGroupToRoomDialog/AddGroupToRoomDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+            onClick={onClose}
+            onFocus={(event: React.FocusEvent) => event.stopPropagation()}
+            margin={{ right: 'small' }}
+          />
+          <LoadingButton
+            label={intl.formatMessage(messages.addGroupsLabel, { numberSelected: selected.length })}
+            primary
+            onClick={() =>
+              mutate({ roomSlug, groupIds: selected.map((index) => groups?.[index]?.id as string) })
+            }
```
kernicPanel, line 137: I think you should put this in a function.

src/frontend/magnify/src/components/rooms/AddGroupToRoomDialog/AddGroupToRoomDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+  return (
+    <Button
+      key={group.id}
+      onClick={(event: React.MouseEvent) => {
+        event.preventDefault();
+        event.stopPropagation();
+        onRemove(group.id);
+      }}
+      href="#"
+      onFocus={(event: React.FocusEvent) => event.stopPropagation()}
```
kernicPanel, line 46: Several questions here. Why do we need:
- to stop the click and focus event propagation
- to specify this href

src/frontend/magnify/src/components/rooms/AddGroupToRoomDialog/AddGroupToRoomDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+            value={
+              <Box wrap direction="row">
+                {selected && selected.length > 0 ? (
+                  selected.map((groupIndex) => renderGroup(groups?.[groupIndex], handleUnselect))
+                ) : (
+                  <Box pad="small">{intl.formatMessage(messages.selectGroupsToAdd)}</Box>
+                )}
+              </Box>
+            }
+            options={(groups || []).map((g) => g.name)}
```
kernicPanel, line 118:
```suggestion
            options={(groups || []).map((group) => group.name)}
```
Please don't use such shortened variable names.

src/frontend/magnify/src/components/rooms/AddGroupToRoomDialog/AddGroupToRoomDialog.tsx
Leonils: Indeed.

src/frontend/magnify/src/components/rooms/AddGroupToRoomDialog/AddGroupToRoomDialog.tsx
Leonils: This is the suggestion from grommet. 

As I understand it, we need to capture these events to stop the propagation to the select component (that would close the selector)

If we don't specify the href, it's considered a "button", and a "button" shouldn't be child of another button, but the select component of grommet is a button. Setting the "href" make it a link, and then it's valid.

You can see the example from grommet [here](https://storybook.grommet.io/?path=/story/input-select-children--children)

Would you have an alternative ?

src/frontend/magnify/src/components/rooms/AddGroupToRoomDialog/AddGroupToRoomDialog.tsx
Leonils: Fair enough

kernicPanel (approved):
:+1:

PR #57, opened 2022-07-08
🐛(swagger) fix api url and auth type

DarkPitoune:
The swagger can be used as a postman if setup accordingly.

It's a fix that was made in a closed PR so I forgot about it. 

The URL of the backend is set in the settings and the auth is set as jwt.


sandbox/settings.py
```diff
@@ -201,6 +201,14 @@ class Base(DRFMixin, MagnifyCoreConfigurationMixin, Configuration):
         "dockerflow.django.middleware.DockerflowMiddleware",
     )
 
+    # Swagger
+    API_URL = values.Value("http://localhost:8071")
```
sampaccoud, line 205: I think you should not put this default value in Base settings but instead you can force its value to "http://localhost:8071" in the Development settings below.

sampaccoud, line 205: and so here do:
```suggestion
    API_URL = values.Value()
```

sampaccoud: from what I see you are overwriting the API_URL value just after creating it so it's the same as before. Did you trying what I suggested and is there a difficulty that I overlooked?

DarkPitoune: Yup it wasn't working anyway ^^'. I tried another way around

sandbox/settings.py
```diff
@@ -201,6 +201,14 @@ class Base(DRFMixin, MagnifyCoreConfigurationMixin, Configuration):
         "dockerflow.django.middleware.DockerflowMiddleware",
     )
 
+    # Swagger
+    API_URL = values.Value("bob")
```
sampaccoud, line 205: debug fail
```suggestion
    API_URL = values.Value()
```

sampaccoud (commented):
Seems good if it works :clap: 
Can you squash the commits into one?

DarkPitoune:
There is only one commit
Am waiting for review before merge

sampaccoud (approved):

PR #59, opened 2022-07-08
🎨(backend) use REST style api path for RoomUser and RoomGroup

kernicPanel:
## Purpose

We want to use api urls in the form `/api/rooms/{room.id}/users/{user.id}/`

## Proposal

- [x] Replace RoomUserViewSet by actions in RoomViewSet.
- [x] Replace RoomGroupViewSet by actions in RoomViewSet.

### DIFF
--- a/src/magnify/apps/core/api.py
+++ b/src/magnify/apps/core/api.py
@@ -12,12 +12,7 @@ from rest_framework.response import Response
 from rest_framework.views import exception_handler as drf_exception_handler
 
 from .models import Room, RoomGroup, RoomUser, User
-from .permissions import (
-    IsObjectAdministrator,
-    IsRelatedRoomAdministrator,
-    IsRoomAdministrator,
-    IsSelf,
-)
+from .permissions import IsObjectAdministrator, IsRoomAdministrator, IsSelf
 from .serializers import (
     PasswordChangeSerializer,
     RegistrationSerializer,
@@ -166,22 +161,6 @@ class UserViewSet(
         return Response(status=204)
 
 
-class RoomGroupViewSet(
-    mixins.CreateModelMixin,
-    mixins.DestroyModelMixin,
-    mixins.RetrieveModelMixin,
-    mixins.UpdateModelMixin,
-    viewsets.GenericViewSet,
-):
-    """
-    API endpoint to access and perform actions on room/group relations.
-    """
-
-    permission_classes = [IsRelatedRoomAdministrator]
-    queryset = RoomGroup.objects.all()
-    serializer_class = RoomGroupSerializer
-
-
 class RoomViewSet(
     mixins.CreateModelMixin,
     mixins.DestroyModelMixin,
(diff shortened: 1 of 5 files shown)


src/magnify/apps/core/api.py
```diff
@@ -234,3 +202,125 @@ def perform_create(self, serializer):
...
+        methods=["get", "put", "delete"],
+        detail=True,
+        url_path="users/(?P<user_pk>[^/.]+)",
+        permission_classes=[IsRoomAdministrator],
+    )
+    # pylint: disable=unused-argument,invalid-name
+    def user(self, request, user_pk, pk=None):
+        """Get, update, or delete a user of a room."""
+        room = self.get_object()
+        room_user = RoomUser.objects.get(room=room, user__pk=user_pk)
```
sampaccoud, line 240: now I know why it should be called `RoomUserRelation`: it took me 1 minute to understand that this was not a user... The reader a RoomUser is a user in a room...

src/magnify/apps/core/api.py
```diff
@@ -234,3 +202,125 @@ def perform_create(self, serializer):
...
+
+        if request.method == "PUT":
+            user = User.objects.get(id=user_pk)
+            serializer = RoomUserSerializer(
+                room_user,
+                data={
+                    "room": room.id,
+                    "user": user.id,
+                    "is_administrator": request.data.get("is_administrator"),
+                },
```
sampaccoud, line 258: In fact it seems you are getting the user by its pk only to get its id, and the room only to get its id. 
```suggestion
            serializer = RoomUserSerializer(
                room_user,
                data={
                    "room": pk,
                    "user": user_pk,
                    "is_administrator": request.data.get("is_administrator"),
                },
```

src/magnify/apps/core/api.py
```diff
@@ -234,3 +202,125 @@ def perform_create(self, serializer):
...
+    @action(
+        methods=["get", "put", "delete"],
+        detail=True,
+        url_path="users/(?P<user_pk>[^/.]+)",
+        permission_classes=[IsRoomAdministrator],
+    )
+    # pylint: disable=unused-argument,invalid-name
+    def user(self, request, user_pk, pk=None):
+        """Get, update, or delete a user of a room."""
+        room = self.get_object()
```
sampaccoud, line 239: I think you don't need this line (you can directly use pk for the PUT below)

src/magnify/apps/core/api.py
```diff
@@ -234,3 +202,125 @@ def perform_create(self, serializer):
...
+    # pylint: disable=unused-argument,invalid-name
+    def users(self, request, pk=None):
+        """Adds a user in a room."""
+        room = self.get_object()
+        user = User.objects.get(id=request.data.get("user"))
+        serializer = RoomUserSerializer(
+            data={
+                "room": room.id,
+                "user": user.id,
+                "is_administrator": request.data.get("is_administrator"),
```
sampaccoud, line 221:
```suggestion
        serializer = RoomUserSerializer(
            data={
                "room": pk,
                "user": request.data.get("user"),
                "is_administrator": request.data.get("is_administrator"),
```

DarkPitoune (approved):

src/magnify/apps/core/api.py
kernicPanel: Relation still confuses me… Maybe RoomUserRole ?

src/magnify/apps/core/api.py
kernicPanel: For the user, I agree!
Four the room, too, but apparently we have to call `self.get_object()` for django to check for `has_object_permission`. And calling it directly will require us to get it anyway.

src/magnify/apps/core/api.py
kernicPanel: Same here.

src/magnify/apps/core/api.py
kernicPanel: And here.

src/magnify/apps/core/api.py
sampaccoud: or `RoomUserAccess` like in [marsha](https://github.com/openfun/marsha)?

src/magnify/apps/core/api.py
sampaccoud: Ok good point!

src/magnify/apps/core/api.py
kernicPanel: Yes, good idea.
Let's do this in a dedicated PR.

PR #68, opened 2022-07-11
Create jitsi view

Leonils:
## Purpose

This create a view for the jitsi iframe, including a disambiguation screen at the beginning if there are more than 1 meeting in progress

Screenshot from 2022-07-11 17-57-11
Screenshot from 2022-07-11 17-56-49

### DIFF
--- a/src/frontend/demo/src/App.tsx
+++ b/src/frontend/demo/src/App.tsx
@@ -8,6 +8,7 @@ import ProfileView from './views/profile';
 import RoomView from './views/room';
 import RoomsView from './views/rooms';
 import RoomSettingsView from './views/roomSettings';
+import JitsiView from './views/jitsi';
 
 export default function App() {
   const { connexionStatus } = useStore();
@@ -35,6 +36,8 @@ export default function App() {
         <Route path="/rooms" element={<RoomsView />} />
         <Route path="/rooms/:slug" element={<RoomView />} />
         <Route path="/rooms/:slug/settings" element={<RoomSettingsView />} />
+        <Route path="/j/m/:meetingId" element={<JitsiView />} />
+        <Route path="/j/:roomSlug" element={<JitsiView />} />
         <Route path="*" element={<NotFoundView />} />
       </Routes>
     </Box>
--- a/src/frontend/demo/src/index.tsx
+++ b/src/frontend/demo/src/index.tsx
@@ -32,7 +32,7 @@ async function render() {
 
   // Create the controller
   // const controller = new DefaultController({ url: 'http://localhost:3000' });
-  const controller = new LogController();
+  const controller = new LogController(process.env.REACT_APP_TEST_TOKEN as string);
   // controller.refreshActivated = false;
   const queryClient = new QueryClient({
     defaultOptions: {
--- a/src/frontend/demo/.env.example
+++ b/src/frontend/demo/.env.example
@@ -1,0 +1,1 @@
+TEST_TOKEN="a token to grant access to any room"
--- a/src/frontend/demo/jest.config.js
+++ b/src/frontend/demo/jest.config.js
@@ -3,13 +3,14 @@
 module.exports = {
   verbose: true,
   transform: {
-    '^.+\\.tsx?$': `ts-jest`,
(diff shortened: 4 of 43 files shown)


kernicPanel:
Could you please rebase ? The conflicts seems easy to solve.

src/frontend/magnify/test.txt
```diff
@@ -0,0 +1,233 @@
+yarn run v1.22.19
```
kernicPanel, line 1: Should we keep this file?

src/frontend/demo/src/index.tsx
```diff
@@ -32,7 +32,8 @@ async function render() {
 
   // Create the controller
   // const controller = new DefaultController({ url: 'http://localhost:3000' });
-  const controller = new LogController();
+  console.log(process.env, process.env.REACT_APP_TEST_TOKEN);
```
kernicPanel, line 35: Don't forget to remove this.

src/frontend/demo/src/App.tsx
```diff
@@ -33,6 +34,8 @@ export default function App() {
         <Route path="/groups" element={<GroupsView />} />
         <Route path="/rooms" element={<RoomsView />} />
         <Route path="/rooms/:slug" element={<RoomView />} />
+        <Route path="/j/m/:meetingId" element={<JitsiView />} />
+        <Route path="/j/:roomSlug" element={<JitsiView />} />
```
kernicPanel, line 38: Why such short path ?

kernicPanel (commented):
I feel that the usage of `disambiguation` a bit convoluted…
Why not simply `selection` ie: `MeetingSelection` ?

src/frontend/demo/src/App.tsx
Leonils: In my opinion, to keep the url as simple as possible (meeting.education/j/my-amazing-room) but still avoiding potential conflicts, so we don't need to forbid some room names because it's the name of other rooms

src/frontend/magnify/test.txt
Leonils: No, indeed

Leonils:
About the Disambiguation, it should not be visible too often, but only when we have a doubt on what the user want to do, for instance if there are 2 meetings at the same time, of very close, so the name. But if you prefer, I can rename it.

Leonils:
Something else @kernicPanel, Could you please check the logs of the test step? The react sdk for jitsi is not transpiled, and that makes the tests to crash on the demo. Have you an idea on how to solve that ? Thank you

kernicPanel:
> About the Disambiguation, it should not be visible too often, but only when we have a doubt on what the user want to do, for instance if there are 2 meetings at the same time, of very close, so the name. But if you prefer, I can rename it.

Ok, let's keep it like this, as clock is ticking :timer_clock:

kernicPanel:
> Something else @kernicPanel, Could you please check the logs of the test step? The react sdk for jitsi is not transpiled, and that makes the tests to crash on the demo. Have you an idea on how to solve that ? Thank you

@jbpenrath found the solution !
In `src/frontend/demo/jest.config.js`:
```
// import type { Config } from "@jest/types";

module.exports = {
  verbose: true,
  transform: {
    '^.+\\.(ts|tsx|js|jsx)$': 'ts-jest',
  },
  moduleNameMapper: {
    '\\.(css|less|scss|sss|styl)$': 'jest-css-modules',
    '@jitsi-magnify/core': `<rootDir>/../magnify/src/index.ts`,
  },
  testEnvironment: 'jsdom',
  setupFilesAfterEnv: ['./jest-setup.ts'],
  moduleDirectories: ['node_modules', 'src'],
  transformIgnorePatterns: ['node_modules/(?!@jitsi/react-sdk)'],
};
```

src/frontend/demo/src/App.tsx
Leonils: Should we close that or would you prefer me to update the path. What are your suggestions ? Thank you

Leonils:
Thank you a lot. It works indeed.

kernicPanel (approved):
:clap:

PR #105, opened 2022-10-04
Add field to room serializer to indicate if the logged-in user is administrator

sampaccoud:
## Purpose

The backend can easily add a field to the room API endpoint to indicate if the logged-in user is an administrator of the room.

In the frontend, we would have to go through all the users in the room to know if the logged-in user is an admin or not.

## Proposal

- [x] Add a field `has_admin_access` to the room serializer
- [x] Correct tests

We took the opportunity to slightly improve existing tests on the room API.


src/magnify/apps/core/serializers/rooms.py
```diff
@@ -104,5 +104,6 @@ def to_representation(self, instance):
                     "groups": groups_serializer.data,
                 }
             )
+        output["has_admin_access"] = is_admin
```
kernicPanel, line 107: Why not keep the same name ?
```suggestion
        output["is_admin"] = is_admin
```

kernicPanel (approved):
Just a suggestion, otherwise :+1:

src/magnify/apps/core/serializers/rooms.py
sampaccoud: My thinking was that `is_admin` is used on the `User` object to indicate that a given user is admin.
What we are indicating here is that the logged in user is admin OF the room on which we put this field... quite different right?

That's why I came out with a different name which, I agree, could be better but that's all I found :man_shrugging:  
Any idea with the context explained above?

kernicPanel: Oh, ok I think I get it : this is a room attribute, not a user one, right ?
If so, maybe something like
```suggestion
        output["administrable"] = is_admin
```

kernicPanel: I had the same feeling :sweat_smile:

PR #109, opened 2022-10-17
✨(api) make username editable via the user API

sampaccoud:
## Purpose

We want to allow changing one's username. This is acceptable as long as the field remains unique.

## Proposal

Remove `username` from the list of readonly fields


src/magnify/apps/core/models.py
```diff
@@ -52,7 +52,7 @@ class User(BaseModel, AbstractBaseUser, PermissionsMixin):
     """
 
     username_validator = RegexValidator(
-        "^[a-zA-Z0-9_-]*$",
+        "^[a-z0-9_-]*$",
         message=_(
             "Username must contain only letters, numbers, underscores and hyphens."
```
kernicPanel, line 57:
```suggestion
            "Username must contain only lower case letters, numbers, underscores and hyphens."
```

kernicPanel (commented):
Just a suggestion regarding helper text, otherwise LGTM

src/magnify/apps/core/models.py
sampaccoud: Good idea :pray:

kernicPanel (commented):
IMHO, a migration should downcase existing usernames in the database, or if no user is in production yet, the existing migration should be updated.

kernicPanel:
> IMHO, a migration should downcase existing usernames in the database, or if no user is in production yet, the existing migration should be updated.

Sorry, I didn't think about it before

sampaccoud:
> IMHO, a migration should downcase existing usernames in the database, or if no user is in production yet, the existing migration should be updated.

Good point ! But this project is not released yet so we are still in the phase where we break the database in each PR!

kernicPanel (approved):
Don't forget to update the migrations with the new text helper.

PR #116, opened 2022-10-25
♻️(frontend) move auth token in localStorage

PanchoutNathan:
## Purpose

sessionStorage only has current tab as context. We move into localStorage

## Proposal

- [x] update http.service.ts
- [x] update users.service.ts


NathanVss (approved):
LGTM :)

PR #118, opened 2022-10-27
♻️(api) remove room token view and accept unregistered rooms on API

sampaccoud:
## Purpose

The existing room token view was delivering Jitsi accesses even on rooms that are now reserved in Magnify. We only want to allow a room after checking if it is reserved. 

## Proposal

We decided to add this feature on the existing room API detail endpoint and handle unregistered rooms on a 404 error when no existing room is found.


src/magnify/apps/core/serializers/rooms.py
```diff
@@ -106,4 +92,18 @@ def to_representation(self, instance):
...
         output["is_administrable"] = is_admin
 
+        if instance.is_public or (
+            user.is_authenticated
+            and (
+                instance.user_accesses.filter(user=user)
+                or instance.group_accesses.filter(
+                    Q(group__members=user) | Q(group__administrators=user)
+                )
+            )
```
carofun, line 102:
```suggestion
            user.is_authenticated
            and instance.has_access)
```
it's only be moved, but it could be handy to add directly a method in the Room model so it can be reused, for instance
def has_access(user):

carofun (approved):
Maybe we could split the commit in 2, one to retrieve the room and the second one about the setting of the moderator value and the use of is_admin in the token, I let it up to you

tests/apps/core/test_core_api_rooms.py
```diff
@@ -114,12 +119,45 @@ def test_api_rooms_retrieve_anonymous_private_slug(self):
             },
         )
 
+    @mock.patch("magnify.apps.core.utils.generate_token", return_value="the token")
```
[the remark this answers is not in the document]
sampaccoud: good point :pray:

src/magnify/apps/core/serializers/rooms.py
sampaccoud: Good idea :star_struck: :pray:

sampaccoud:
> Maybe we could split the commit in 2, one to retrieve the room and the second one about the setting of the moderator value and the use of is_admin in the token, I let it up to you

Looked like a good idea, but in fact the "moderator" refacto is intricated to the fact that we want to generate a token for unregistered rooms.... so it makes sense to keep it in the same commit :man_shrugging:

PR #119, opened 2022-10-27
🐛(models) add default language to user

sampaccoud:
## Purpose

The `createsuperuser` management command won't work if the language field is required and not set by default.

## Proposal

Add `LANGUAGE_CODE` as default language


tests/sandbox/test_management_commands.py
```diff
@@ -0,0 +1,31 @@
...
+Test suite for magnify's main module
+"""
+from io import StringIO
+from unittest import mock
+
+from django.core.management import call_command
+from django.test import TestCase
+
+# sandbox
+from sandbox.settings import get_release
```
jbpenrath, line 11: This is unused.
```suggestion
```

tests/sandbox/test_management_commands.py
```diff
@@ -0,0 +1,31 @@
+"""
+Test suite for magnify's main module
+"""
+from io import StringIO
+from unittest import mock
```
jbpenrath, line 5: This is unused.
```suggestion
```

src/magnify/apps/core/models.py
```diff
@@ -76,6 +76,7 @@ class User(BaseModel, AbstractBaseUser, PermissionsMixin):
     language = models.CharField(
         max_length=10,
         choices=lazy(lambda: settings.LANGUAGES, tuple)(),
+        default=settings.LANGUAGE_CODE,
```
jbpenrath, line 79: If nothing is yet in production, maybe you could recreate the initial migration?

sampaccoud: True

jbpenrath (approved):
LGTM ✅

PR #121, opened 2022-10-28
Update models.py

ptitloup:
Set weekdays to None if recurrence is not equal to Weekly

### DIFF
--- a/tests/apps/core/test_core_models_meetings_get_occurences.py
+++ b/tests/apps/core/test_core_models_meetings_get_occurences.py
@@ -200,6 +200,27 @@ class OccurencesMeetingsModelsTestCase(TestCase):
             ["Weekdays should contain the start date."],
         )
 
+    def test_models_meetings_get_occurrences_weekly_reset_weekdays(self):
+        """reset weekdays if recurrence not equal to weekly"""
+        meeting = MeetingFactory(
+            start=date(2022, 7, 6),
+            recurrence="weekly",
+            frequency=1,
+            recurring_until=date(2022, 8, 2),
+            nb_occurrences=None,
+            meeting.weekdays = "26"
+        )
+        self.assertEqual(meeting.weekdays, "26")
+        meeting.recurrence = "daily"
+        meeting.save()
+        meeting.refresh_from_db()
+        self.assertEqual(meeting.weekdays, None)
+        meeting.recurrence = "weekly"
+        meeting.weekdays = "126"
+        meeting.save()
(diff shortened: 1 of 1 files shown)


sampaccoud:
> Set weekdays to None if recurrence is not equal to Weekly

Nice to see you here  @ptitloup :star_struck:
Can you help me understand which case goes wrong when we don't do what?

sampaccoud:
I was thinking of what test I should write to evidence the problem but I'm a bit far from the code after the summer :sweat_smile: 
My understanding at the moment is that, if you switched to monthly recurrence, the `weekday` field would keep its value but it would not affect your monthly recurrence. Right?
So it's more correct to reset the field but there is no misbehaviour associated or am I missing something and there is a test to write to demonstrate a bug?

ptitloup:
> My understanding at the moment is that, if you switched to monthly recurrence, the weekday field would keep its value but it would not affect your monthly recurrence. Right?
Yep !
> So it's more correct to reset the field but there is no misbehavior associated or am I missing something and there is a test to write to demonstrate a bug?
Correct

ptitloup:
Hello, I made some unit tests to check reset weekdays. You can do something like that : 

`    
def test_models_meetings_get_occurrences_weekly_reset_weekdays(self):
        """reset weekdays if recurrence not equal to weekly"""
       meeting = MeetingFactory(
            start=date(2022, 7, 6),
            recurrence="weekly",
            frequency=1,
            recurring_until=date(2022, 8, 2),
            nb_occurrences=None,
            meeting.weekdays = "26"
        )

        self.assertEqual(meeting.weekdays, "26")
        meeting.recurrence = "daily"
        meeting.save()
        meeting.refresh_from_db()

        self.assertEqual(meeting.weekdays, None)
        meeting.recurrence = "weekly"
        meeting.weekdays = "126"
        meeting.save()
        meeting.refresh_from_db()
        self.assertEqual(meeting.weekdays, "126")
`

sampaccoud:
@ptitloup thanks. Do you want to add it to the PR and rebase it so we merge it?

ptitloup:
> @ptitloup thanks. Do you want to add it to the PR and rebase it so we merge it?

Ok ! I'll do that asap

ptitloup:
Done !

sampaccoud (approved):

sampaccoud:
:rocket: Let's merge :wink:  Thanks for your contribution

ptitloup:
my pleasure !

PR #123, opened 2022-10-31
🐛(api) normalize room names in API calls

sampaccoud:
## Purpose

Calls to the room detail API endpoint may contain accents and upper case letters in the room name. We want to normalize them to the slug we save in room names.
Said differently "Réunion" and "reunion" should be the same room.

## Proposal

- [x] Call slugify on the room name before looking for a room with this slug
- [x] Add tests to secure this behavior

### DIFF
--- a/tests/apps/core/test_core_api_rooms.py
+++ b/tests/apps/core/test_core_api_rooms.py
@@ -119,6 +119,23 @@ class RoomsApiTestCase(APITestCase):
             },
         )
 
+    def test_api_rooms_retrieve_anonymous_private_slug_not_normalized(self):
+        """Getting a room by a slug that is not normalized should work."""
+        room = RoomFactory(name="Réunion", is_public=False)
+        response = self.client.get("/api/rooms/Réunion/")
+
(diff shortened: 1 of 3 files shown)


tests/apps/core/test_core_api_rooms.py
```diff
@@ -143,6 +160,27 @@ def test_api_rooms_retrieve_anonymous_unregistered_allowed(self, mock_token):
             AnonymousUser(), "unregistered-room", is_admin=True
         )
 
+    @override_settings(ALLOW_UNREGISTERED_ROOMS=True)
+    @mock.patch("magnify.apps.core.utils.generate_token", return_value="the token")
+    def test_api_rooms_retrieve_anonymous_unregistered_allowed_not_normalized(self, mock_token):
+        """Getting a room by a slug that is not normalized should work."""
```
carofun, line 166:
```suggestion
        """Getting an unregistered room by a slug that is not normalized should work and create a room with the slugged name."""
```

carofun (approved):

PR #124, opened 2022-11-04
🌐(frontend) Compile translation demo

PanchoutNathan:
## Purpose

Compile demo and package translations into one file

## Proposal

- [x] Edit demo package.json for compile-translations command


src/frontend/demo/package.json
```diff
@@ -50,7 +50,7 @@
     "typescript": "4.7.3"
   },
   "scripts": {
-    "compile-translations": "./i18n/compile-translations.js",
```
jbpenrath, line 53: I think you should explain in your commit that using this script from core package also compile message from core package into demo translations and it allows translation override.

src/frontend/magnify/src/components/app/MagnifyTestingProvider/index.tsx
```diff
@@ -33,7 +33,7 @@ const locale = 'en-US';
```
jbpenrath, line 33: I think your commit message did not pass `lint-git` because commit object starts with an uppercased letter.

```diff
- 🐛(frontend) Translation of YUP messages
+ 🐛(frontend) translation of YUP messages
```

src/frontend/magnify/src/components/auth/SignupForm/SignupForm.tsx
```diff
@@ -125,15 +126,15 @@ export default function SignupForm() {
...
+  }, [locale.currentLocale]);
 
   const initialValues: SignupFormValues = {
     name: '',
     email: '',
     username: '',
     password: '',
     confirmPassword: '',
-    language: MagnifyLocales.EN,
+    language: locale.currentLocale ?? MagnifyLocales.EN,
```
jbpenrath, line 137: Hmm.. `currentLocale` cannot be `null` or `undefined` as TranslationProvider defaults to `MagnifyLocales.EN`

src/frontend/magnify/src/components/design-system/Formik/LanguageChange/index.tsx
```diff
@@ -0,0 +1,14 @@
+import { useFormikContext } from 'formik';
+import * as React from 'react';
+import { PropsWithChildren, useEffect } from 'react';
+import { useLocale } from '../../../../i18n/TranslationProvider/TranslationsProvider';
+
+export const FormikLanguageChange = ({ children }: PropsWithChildren) => {
```
jbpenrath, line 6: What do you think to call this component `LocalizedForm`  then instead of simply return `children` it returns `<Form>{children}</Form>` ?

src/frontend/magnify/src/i18n/FormErrors/index.tsx
```diff
@@ -1,21 +1,17 @@
 import * as React from 'react';
-import { useEffect } from 'react';
 import { useIntl } from 'react-intl';
 import { setLocale } from 'yup';
 import { validationMessages } from '../Messages';
 
 export interface FormErrorsProps {}
 
-export function FormErrors(props: FormErrorsProps) {
+export function FormErrors() {
```
jbpenrath, line 8: Can you add a comment to explain the purpose of this component and how and where it should be used. I do not understand at first.

src/frontend/magnify/src/i18n/FormErrors/index.tsx
```diff
@@ -1,21 +1,17 @@
...
   const intl = useIntl();
+  setLocale({
+    mixed: {
+      required: intl.formatMessage(validationMessages.required),
+    },
+  });
```
jbpenrath, line 14: Cannot be triggered only when `intl.locale` change ?

src/frontend/demo/package.json
PanchoutNathan: Agreed

src/frontend/magnify/src/components/auth/SignupForm/SignupForm.tsx
PanchoutNathan: of course

src/frontend/magnify/src/i18n/FormErrors/index.tsx
PanchoutNathan: Done

src/frontend/magnify/src/components/profile/IdentityForm/IdentityForm.test.tsx
```diff
@@ -21,7 +22,7 @@ describe('IdentityForm', () => {
     await user.clear(nameInput);
     fireEvent.blur(nameInput);
 
-    await screen.findByText('name is a required field');
+    await screen.findByText(validationMessages.required.defaultMessage);
```
jbpenrath, line 25: In this case, we prefer to use a static value. First it helps to better understand the context of this test. Then it warns us about regression related to the label.

jbpenrath (approved):
Update the test then merge it !

PR #125, opened 2022-11-04
✨(frontend) set current lang request headers

PanchoutNathan:
## Purpose

Currently, the messages sent by the back are conditioned by the language of the browser. To make the messages sent by the back correspond to the language chosen by the user and not to the language of the browser we update the headers

## Proposal

- [x] init headers with the locale in localStorage
- [x] update headers when we change user locale

### DIFF
--- a/src/frontend/magnify/src/services/http/http.service.ts
+++ b/src/frontend/magnify/src/services/http/http.service.ts
@@ -1,6 +1,7 @@
 import axios, { AxiosError } from 'axios';
+
 import { RefreshTokenResponse } from '../../types';
-import { UsersApiRoutes } from '../../utils';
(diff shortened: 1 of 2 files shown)


src/frontend/magnify/src/services/http/http.service.ts
```diff
@@ -81,3 +89,7 @@ MagnifyApi.interceptors.response.use(
     return Promise.reject(error);
   },
 );
+
+setTimeout(() => {
+  HttpService.setRequestLanguage(getLocaleStorageLang());
+}, 0);
```
jbpenrath, line 95: Why do you need to do this ? We could set a default headers like you did to set the `Authorization` header.
Why not use an IIFE instead of this "instant" timeout.

PR #126, opened 2022-11-04
♻️(frontend) update account page

PanchoutNathan:
## Purpose

Some features such as avatar upload or account deletion are not yet available. They are therefore not displayed to the user.

## Proposal

- [x] remove avatar picker
- [x] remove delete account block

### DIFF
--- a/src/frontend/demo/src/views/profile/account/index.tsx
+++ b/src/frontend/demo/src/views/profile/account/index.tsx
@@ -1,9 +1,4 @@
-import {
-  DeleteAccountBlock,
-  IdentityBlock,
-  PasswordUpdateBlock,
-  useTranslations,
-} from '@jitsi-magnify/core';
+import { IdentityBlock, PasswordUpdateBlock, useTranslations } from '@jitsi-magnify/core';
 import * as React from 'react';
 import { defineMessages } from 'react-intl';
 import { DefaultPage } from '../../../components/DefaultPage';
@@ -22,7 +17,6 @@ export function ProfileAccountView() {
     <DefaultPage title={intl.formatMessage(messages.profileViewTitle)}>
       <IdentityBlock />
       <PasswordUpdateBlock />
-      <DeleteAccountBlock />
     </DefaultPage>
   );
 }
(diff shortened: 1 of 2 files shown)


src/frontend/magnify/src/i18n/TranslationProvider/TranslationsProvider.tsx
```diff
@@ -1,6 +1,7 @@
 import { Box, Spinner } from 'grommet';
```
jbpenrath, line 1: You have to drop this commit

src/frontend/demo/src/views/profile/account/index.tsx
```diff
@@ -1,9 +1,4 @@
-import {
```
jbpenrath, line 1: There is a 🔥 gitmoji to type commit which deletes code.

src/frontend/magnify/src/components/profile/IdentityBlock/IdentityBlock.tsx
```diff
@@ -17,9 +16,9 @@ export default function IdentityBlock({ margin = { vertical: 'small' } }: Identi
   return (
     <Card background="white" margin={margin} pad={isSmall ? 'medium' : 'small'}>
       <Box align={'center'} direction={isSmall ? 'column' : 'row'} justify={'center'}>
-        <Box margin={{ vertical: 'auto', horizontal: 'medium' }}>
-          <AvatarForm key={user?.id || ''} id={user?.id} />
-        </Box>
+        {/*<Box margin={{ vertical: 'auto', horizontal: 'medium' }}>*/}
+        {/*  <AvatarForm key={user?.id || ''} id={user?.id} />*/}
+        {/*</Box>*/}
```
jbpenrath, line 21: Hmmm... I assume this code is removed temporarily but why not remove it completely then you will be able to find this commit from this commit later, WDYT ?

If you prefer to let this code commented, that's ok but add a comment to explain why this code is currently disabled.

src/frontend/magnify/src/i18n/TranslationProvider/TranslationsProvider.tsx
PanchoutNathan: Resolve when merge the root branch

PR #131, opened 2022-11-23
Update models.py

ptitloup:
Add method to get the next occurrence of meeting from today.

## Purpose

We need to know when is the next occurrence of a meeting from today. The method next_occurrence need of occurrence date of a meeting and except the start date we don't any one. The get occurrence method need a end date and we don't have one except the recurring until date. 

## Proposal

We don't know if it's a good method but we want to purpose you this solution to have your feedback about that. The method check if today is an occurrence of meeting and if not parse next_occurence meeting until find one

### DIFF
--- a/src/magnify/apps/core/models.py
+++ b/src/magnify/apps/core/models.py
@@ -430,7 +430,22 @@ class Meeting(BaseModel):
             self.reset_recurrence()
 
         super().save(*args, **kwargs)
-
+ 
+    def next_occurrence_from_today(self):
+        """
+        Returns the next occurrence of the meeting from today.
+        """
+        if self.start == timezone.now().date():
+            start_datetime = dt.combine(self.start, self.start_time)
+            start_datetime = timezone.make_aware(start_datetime)
+            start_datetime = start_datetime + self.expected_duration
+            if start_datetime > timezone.now():
+                return self.start
+        next_one = self.next_occurrence(self.start)
+        while next_one < timezone.now().date():
+            next_one = self.next_occurrence(next_one)
+        return next_one
+ 
     def next_occurrence(self, current_date):
         """
         This method takes as assumption that the current date passed in argument


src/magnify/apps/core/models.py
```diff
@@ -430,7 +430,22 @@ def save(self, *args, **kwargs):
             self.reset_recurrence()
 
         super().save(*args, **kwargs)
-
+ 
+    def next_occurrence_from_today(self):
+        """
+        Returns the next occurrence of the meeting from today.
+        """
+        if self.start == timezone.now().date():
```
sampaccoud, line 438: First I think we need to add a "timezone" field to the model to be aware of what the user meant when s.he entered the date and time of the meeting? (cf google... it is set by default from your profile but you can modify it when scheduling your meeting...)

Then we will be able to compute the exact and aware datetime of the meeting in order to compare it with timezone.now()?

sampaccoud: maybe we should compare with start + 1/2*expected_duration to know if the meeting is passed or not :smile:

ptitloup: I use the "timezone" object to pay attention to the user's time zone as advised by django. Maybe we could use datetimefield instead of start and start_time

ptitloup: > maybe we should compare with start + 1/2*expected_duration to know if the meeting is passed or not 😄

Why not... I used start + expected_duration to know if the meeting is still active or not

sampaccoud: Yes, that's the only way to know the exact time we are talking about :thinking: I remember having a reason not to merge these 2 infos into one field but I forgot and now it seems stupid to me. So yes I think we have to do that!

sampaccoud: I will prototype something next week.

sampaccoud: Not exactly next week but I propose this fix: https://github.com/openfun/jitsi-magnify/pull/140

ptitloup: Hello, thanks for the information. I will have a look to this as soon as possible

PR #132, opened 2022-11-28
Fix room issues on the backend which arose after finalizing the frontend VMR feature

sampaccoud:
## Purpose

A lot of small issues on the backend arose while finalizing the frontend VMR feature. 

## Proposal

- Refactor room accesses to RBAC and open a specific API endpoint. The API endpoints for room accesses were nested under the room endpoint which lead to access control issues and was unnecessarily complicated. We take the opportunity to switch to RBAC so we can differentiate owners from administrators and from simple members of a room.
- Hide Jitsi configuration for non admin users. It includes the conference password when it is set so only administrators should be allowed to see it.
- On the user API endpoint, allow searching users with case insensitive username/email
- Add "distinct" clause to the room API list endpoint query to avoid duplicates
- Allow administrators to modify the public status of a room
- Ensure that only owners of a room can delete it
- Make rooms private by default


tests/apps/core/test_core_api_room_user_access.py
```diff
@@ -163,171 +358,504 @@ def test_api_room_user_accesses_create_anonymous(self):
...
+                {**old_values, field: value},
+                HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+            )
+            if field == "role":
+                self.assertEqual(response.status_code, 403)
+            else:
+                self.assertEqual(response.status_code, 200)
+            access.refresh_from_db()
+            updated_values = RoomUserAccessSerializer(instance=access).data
+            self.assertEqual(updated_values, old_values)
```
carofun, line 738:
```suggestion
        for field, value in new_values.items():
            response = self.client.put(
                f"/api/room-user-accesses/{access.id!s}/",
                {**old_values, field: value},
                HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
            )
            if field == "role":
                self.assertEqual(response.status_code, 403)
            else:
                self.assertEqual(response.status_code, 200)
            access.refresh_from_db()
            updated_values = RoomUserAccessSerializer(instance=access).data
            self.assertEqual(updated_values, old_values)
```
?

src/magnify/apps/core/models.py
```diff
@@ -222,17 +237,26 @@ def jitsi_name(self):
...
     def is_administrator(self, user):
         """Check if a user is administrator of the room."""
-        if not user.is_authenticated:
+        if not user or not user.is_authenticated:
             return False
 
         return (
-            self.user_accesses.filter(is_administrator=True, user=user).exists()
+            self.user_accesses.filter(
+                role__in=[UserRoleChoices.ADMIN, UserRoleChoices.OWNER], user=user
```
carofun, line 245:
```suggestion
                role__in=[UserRoleChoices.ADMIN, UserRoleChoices.OWNER], user=user
```
the question is administrator but there with this he could only be the owner, is this correct? You have the method is_owner as well

src/magnify/apps/core/models.py
```diff
@@ -262,8 +288,32 @@ class Meta:
...
+
+    def save(self, *args, **kwargs):
+        """Make sure we keep at least one owner in a room."""
+        if self.pk and self.role != UserRoleChoices.OWNER:
+            accesses = self._meta.model.objects.filter(
+                room=self.room, role=UserRoleChoices.OWNER
+            ).only("pk")
+            if len(accesses) == 1 and accesses[0].pk == self.pk:
+                raise PermissionDenied("A room should keep at least one owner.")
+        return super().save(*args, **kwargs)
```
carofun, line 304:
```suggestion
        """Make sure we keep at least one owner in a room."""
        if self.pk and self.role != UserRoleChoices.OWNER:
            accesses = self._meta.model.objects.filter(
                room=self.room, role=UserRoleChoices.OWNER
            ).only("pk")
            if len(accesses) == 1 and accesses[0].pk == self.pk:
                raise PermissionDenied("A room should keep at least one owner.")
        return super().save(*args, **kwargs)
```
?

carofun (commented):
minor:
🔒️(api) hide configuration for non admin users 
=> three dots seems unnecessary 

It's easier to keep it as a normal field **is** the serializer and remove => **in** the serializer and remove

src/magnify/apps/core/serializers/rooms.py
```diff
@@ -44,18 +61,42 @@ class RoomUserAccessSerializer(RoomAccessSerializerMixin, serializers.ModelSeria
...
         model = models.RoomUserAccess
-        fields = ["id", "user", "room", "is_administrator"]
+        fields = ["id", "user", "room", "role"]
         read_only_fields = ["id"]
 
+    def update(self, instance, validated_data):
+        """Make user and room field readonly but only on update."""
+        validated_data.pop("room", None)
+        validated_data.pop("user", None)
+        return super().update(instance, validated_data)
```
carofun, line 71: as it is common with RoomGroupAccessSerializer maybe it could inherit from a mixin

tests/apps/core/test_core_api_room_user_access.py
```diff
@@ -12,147 +13,341 @@
...
             HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
         )
-        self.assertEqual(response.status_code, 403)
-
-        self.assertEqual(
-            response.json(),
-            {"detail": "You do not have permission to perform this action."},
-        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.json()), 10)
```
carofun, line 119: ditto as group file, the 10 results are not clear to me compared to the 8 of the test after, maybe we could do some control on the results returned

tests/apps/core/test_core_api_room_user_access.py
```diff
@@ -163,171 +358,504 @@ def test_api_room_user_accesses_create_anonymous(self):
...
+            {
+                "detail": "You must be administrator or owner of a room to add accesses to it."
+            },
+        )
+        self.assertFalse(RoomUserAccess.objects.filter(user=other_user).exists())
+
+    def test_api_room_user_accesses_create_administrators_except_owner_direct(self):
+        """
+        A user who is direct administrator in a room should be allowed to create
+        room user accesses in this room for roles other than owner.
```
carofun, line 417:
```suggestion
        A user who is direct administrator in a room should be allowed to create
        room user accesses in this room for roles other than owner.
```
the test speaks about owner but doesn't show it

tests/apps/core/test_core_api_room_user_access.py
```diff
@@ -163,171 +358,504 @@ def test_api_room_user_accesses_create_anonymous(self):
...
+            {
+                "user": str(other_user.id),
+                "room": str(room.id),
+                "role": random.choice(["member", "administrator"]),
+            },
+            HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+        )
         self.assertEqual(response.status_code, 201)
         self.assertEqual(RoomUserAccess.objects.count(), 2)
         self.assertTrue(RoomUserAccess.objects.filter(user=other_user).exists())
```
carofun, line 436:
```suggestion
 response = self.client.post(
            "/api/room-user-accesses/",
            {
                "user": str(other_user.id),
                "room": str(room.id),
                "role": "owner",
            },
            HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
        )
        self.assertEqual(response.status_code, 403)
        self.assertEqual({"detail":"Only owners of a room can assign other users as owners."}, response.json())
```

tests/apps/core/test_core_api_room_user_access.py
```diff
@@ -163,171 +358,504 @@ def test_api_room_user_accesses_create_anonymous(self):
...
                 "user": str(other_user.id),
-                "is_administrator": is_administrator,
+                "room": str(room.id),
+                "role": random.choice(["member", "administrator"]),
             },
             HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
         )
         self.assertEqual(response.status_code, 201)
         self.assertEqual(RoomUserAccess.objects.count(), 1)
         self.assertTrue(RoomUserAccess.objects.filter(user=other_user).exists())
```
carofun, line 461: ditto for owner

src/magnify/apps/core/serializers/rooms.py
```diff
@@ -1,39 +1,56 @@
...
     """
-    A serializer mix to share controling that the logged-in user submitting a room access object
+    A serializer mixin to share controling that the logged-in user submitting a room access object
     is administrator on the targeted room.
     """
 
+    # pylint: disable=too-many-boolean-expressions
+    def validate(self, data):
+        """
+        Check that start is before finish.
```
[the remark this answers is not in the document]
sampaccoud: Good catch :clap:

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -12,148 +13,331 @@
...
+            HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), [])
+
+    def test_api_room_group_accesses_list_members(self):
         """
-        Anonymous users should not be allowed to retrieve a room group access.
+        Authenticated users should not be allowed to list room group accesses for a room
+        in which they are a simple members, be it directly or via a group.
```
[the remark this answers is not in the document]
sampaccoud: :+1:

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -165,225 +349,485 @@ def test_api_room_group_accesses_create_anonymous(self):
...
         self.assertEqual(RoomGroupAccess.objects.count(), 1)
         self.assertTrue(RoomGroupAccess.objects.filter(group=group).exists())
 
-    def test_api_room_group_accesses_create_administrator_groups(self):
+    def test_api_room_group_accesses_create_administrators_via_group(self):
+        """
+        A user who is administrator via a group in a room should be allowed to create
+        room group accesses in this room for any existing role provided
+        s.he is administrator of the related group.
         """
```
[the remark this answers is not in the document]
sampaccoud: The `if` is not necessary.
The sentence is using the pattern: [I will come to visit you tomorrow, provided it doesn't rain](https://www.wordreference.com/enfr/provided)

Ok for the rest.

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -165,225 +349,485 @@ def test_api_room_group_accesses_create_anonymous(self):
...
+        self.assertEqual(RoomGroupAccess.objects.filter(group=group).count(), 1)
 
-    def test_api_room_group_accesses_update_administrator_users(self):
+    def test_api_room_group_accesses_create_owners(self):
         """
-        Direct administrators of a room should be allowed to update its related
-        room group accesses.
+        A user who is owner in a room should be allowed to create
+        room group accesses in this room for all roles provided
+        s.he is administrator of the related group.
```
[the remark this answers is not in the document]
sampaccoud: not "the owner" but "an owner" as there can be several.
ditto for provided.
ok for the rest.

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -165,225 +349,485 @@ def test_api_room_group_accesses_create_anonymous(self):
...
+                )
+            else:
+                self.assertEqual(updated_values, old_values)
+
+    def test_api_room_group_accesses_update_administrators_via_group(self):
+        """
+        A user who is a administrator in a room, directly or via a group, should be allowed
+        to update a group access for this room even if s.he has no administrator rights
+        on the group (restriction only applies for creating accesses).
+        """
```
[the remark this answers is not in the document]
sampaccoud: :+1:

tests/apps/core/test_core_api_room_user_access.py
```diff
@@ -163,171 +358,504 @@ def test_api_room_user_accesses_create_anonymous(self):
...
+            )
+            self.assertEqual(response.status_code, 403)
+            access.refresh_from_db()
+            updated_values = RoomUserAccessSerializer(instance=access).data
+            self.assertEqual(updated_values, old_values)
+
+    def test_api_room_user_accesses_update_administrator_to_owner(self):
+        """
+        A user who is an administrator in a room, be it directly or via a group, should
+        not be allowed to update the user access of another user to grant her ownership.
```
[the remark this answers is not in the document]
sampaccoud: :+1:

src/magnify/apps/core/models.py
sampaccoud: It's tested in the file `test_core_api_room_user_access.py`. 
The test is called: `test_api_room_user_access_delete_owners_last_owner`

src/magnify/apps/core/serializers/rooms.py
sampaccoud: it's common but not really because it is with different fields... so that's the most straightforward?

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -12,148 +13,331 @@
...
+        RoomGroupAccessFactory(room=private_room)
+        RoomGroupAccessFactory(room=private_room, role="member")
+        RoomGroupAccessFactory(room=private_room, role="administrator")
+
+        response = self.client.get(
+            "/api/room-group-accesses/",
+            HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
         )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.json()), 8)
```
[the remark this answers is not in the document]
sampaccoud: Because there are the 2 group accesses that we created to give access to the user via a group :shrug:

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -12,148 +13,331 @@
...
+                response = self.client.get(
+                    f"/api/room-group-accesses/{access.id!s}/",
+                    HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+                )
+                self.assertEqual(response.status_code, 403)
+                self.assertEqual(
+                    response.json(),
+                    {"detail": "You do not have permission to perform this action."},
+                )
+
```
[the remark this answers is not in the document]
sampaccoud: For complicated factorization yes he was right! I'm just doing simple loops here. Did you get it straightaway or was it disturbing. If it was hard to understand in one glance, i can split it in several tests but it will take a lot of place. Compromises :man_shrugging:

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -165,225 +349,485 @@ def test_api_room_group_accesses_create_anonymous(self):
...
-        response = self.client.put(
-            f"/api/rooms/{access.room.id!s}/groups/{access.group.id!s}/",
+        self.assertEqual(RoomGroupAccess.objects.filter(group=group).count(), 1)
+        response = self.client.post(
+            "/api/room-group-accesses/",
             {
-                "is_administrator": not is_administrator,
+                "group": str(group.id),
+                "room": str(room.id),
+                "role": "owner",
```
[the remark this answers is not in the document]
sampaccoud: I volontarily use the string directly in tests so they are fixed and will need to be updated and checked if the choices are modified in the code.

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -165,225 +349,485 @@ def test_api_room_group_accesses_create_anonymous(self):
...
+            response = self.client.put(
+                f"/api/room-group-accesses/{access.id!s}/",
+                {**old_values, field: value},
+                HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+            )
+
+            self.assertEqual(response.status_code, 200)
+            access.refresh_from_db()
+            updated_values = RoomGroupAccessSerializer(instance=access).data
+            if field == "role":
```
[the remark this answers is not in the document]
sampaccoud: :+1:

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -165,225 +349,485 @@ def test_api_room_group_accesses_create_anonymous(self):
...
         response = self.client.delete(
-            f"/api/rooms/{access.room.id!s}/groups/{access.group.id!s}/",
+            f"/api/room-group-accesses/{access.id!s}/",
             HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
         )
 
         self.assertEqual(response.status_code, 204)
         self.assertEqual(RoomGroupAccess.objects.count(), 1)
         self.assertFalse(RoomGroupAccess.objects.filter(group=access.group).exists())
+
```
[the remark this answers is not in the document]
sampaccoud: Yes! Good point. The idea is that, when adding a group to a room, I only pick among my own groups. But once a group is added to a room with a role, any admin on the room can modifiy the role given to this group and eventually remove it!
I will add this comment to the docstring.

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -165,225 +349,485 @@ def test_api_room_group_accesses_create_anonymous(self):
...
+        access = RoomGroupAccessFactory(room=room)
+
+        jwt_token = AccessToken.for_user(user)
+
+        self.assertEqual(RoomGroupAccess.objects.count(), 1)
+        self.assertTrue(RoomGroupAccess.objects.filter(group=access.group).exists())
+        response = self.client.delete(
+            f"/api/room-group-accesses/{access.id!s}/",
+            HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+        )
```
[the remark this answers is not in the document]
sampaccoud: I will also add my explanation to this test's docstring.

tests/apps/core/test_core_api_room_user_access.py
sampaccoud: yes I'm testing everything else than owner. The owner is tested in the following test. Let me add this precision.

tests/apps/core/test_core_api_room_user_access.py
sampaccoud: I didn't get this comment.

tests/apps/core/test_core_api_room_user_access.py
sampaccoud: I didn't get this comment.

sampaccoud: I added an explanation to the docstring that the owner role is tested in the subsequent test.

src/magnify/apps/core/api/rooms.py
```diff
@@ -91,134 +92,12 @@ def list(self, request, *args, **kwargs):
         return response.Response(serializer.data)
 
     def perform_create(self, serializer):
-        """Set the current user as administrators of the newly created room."""
+        """Set the current user as administrator of the newly created room."""
         room = serializer.save()
```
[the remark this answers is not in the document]
sampaccoud: Good catch!

tests/apps/core/test_core_api_room_group_access.py
```diff
@@ -165,225 +349,485 @@ def test_api_room_group_accesses_create_anonymous(self):
...
+            },
+        )
+        self.assertEqual(RoomGroupAccess.objects.count(), 1)
+
+    def test_api_room_group_accesses_create_administrators_direct(self):
+        """
+        A user who is a direct administrator in a room should be allowed to create
+        room group accesses in this room for any existing role provided
+        s.he is administrator of the related group.
+        """
```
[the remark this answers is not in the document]
sampaccoud: so what can I do to make it clearer?

src/magnify/apps/core/api/rooms.py
```diff
@@ -282,3 +161,90 @@ def meetings(self, request, pk=None):
...
+            .distinct()
+        )
+
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+
+        serializer = self.get_serializer(queryset, many=True)
+        return response.Response(serializer.data)
```
qbey, line 203: Is the reimplementation of the list viewset mandatory? I'm may have missed something, but why don't rely on the DRF methods directly? If you want a mixin:
``` python
class RoomAdministratorListsModelMixin:
    """List mixin for room access API."""

    def get_permissions(self):
        """User only needs to be authenticated to list rooms access"""
        if self.action in ["list"]:
            permission_classes = [UserIsAuthenticated]
        else:
            return super().get_permissions()

        return [permission() for permission in permission_classes]

    def get_queryset(self):
        """Return the queryset according to the action."""
        queryset = super().get_queryset()
        if self.action in ["list""]:
            queryset = (
                .filter(
                    Q(
                        room__user_accesses__user=user,
                        room__user_accesses__role__in=[
                            models.UserRoleChoices.ADMIN,
                            models.UserRoleChoices.OWNER,
                        ],
                    )
                    | Q(
                        room__group_accesses__group__members=user,
                        room__group_accesses__role=models.GroupRoleChoices.ADMIN,
                    )
                )
                .distinct()
            )
        return queryset
```

I'm not sure it would work, but it keeps the concerns separation

``` python
class RoomUserAccessViewSet(
    RoomAdministratorListsModelMixin,
    ModelViewSet,
):
    ...
```

src/magnify/apps/core/permissions.py
```diff
@@ -63,16 +65,24 @@ def has_object_permission(self, request, view, obj):
             return True
 
         user = request.user
+
+        if request.method == "DELETE":
+            return obj.user_accesses.filter(
+                role=UserRoleChoices.OWNER, user=user
+            ).exists()
```
qbey, line 72: I'm not sure it is a good pattern for a permission to manage different permissions. It would be more explicit if a second permission was used at the viewset level according to the action.

qbey (commented):
Nice work, I've only read part of the `(rooms) refactor room accesses to RBAC and open specific API endpoint` commit

src/magnify/apps/core/permissions.py
sampaccoud: I'm not sure either actually! I would say it's nice to handle all permission stuff in this class and have one permission class per specific endpoint? But anyway for certain permissions like field-by-field update permissions, I have to handle it in the serializer...

src/magnify/apps/core/api/rooms.py
sampaccoud: Oh wow... all tests are passing so I'm pretty confident that it works! Much nicer split of responsibility yes! :medal_sports:

src/magnify/apps/core/permissions.py
sampaccoud: After playing a bit with it I think it makes it harder to split permission code between the viewset and the permission class... 
It may be a better pattern when we have simple permission checks like IsStaff, IsAuthenticated etc. that we can compose.
But in the case of RBAC like here it is probably easier to have one permission class per object.... and a lot of tests!!

src/magnify/apps/core/api/rooms.py
sampaccoud: Added a fixup: https://github.com/openfun/jitsi-magnify/pull/132/commits/eac907f402f55f213c50e9b2e3f64d0ecd6d1617

qbey (approved):
Nice job ;)

tests/apps/core/test_core_api_room_user_access.py
sampaccoud: ok understood. This was already handled in a subsequent test.

PR #135, opened 2022-12-12
♻️(api) show room users and groups to members on room detail endpoint

sampaccoud:
## Purpose

Only administrators and owners of a room could see its related users and groups. We think all members of a room should see them.

## Proposal

Add "users" and "groups" fields to the payload returned on the room detail endpoint for a room's members.

I took the opportunity to refactor access rights methods and optimize the number of database query they generate.

### DIFF
--- a/src/magnify/apps/core/models.py
+++ b/src/magnify/apps/core/models.py
@@ -21,19 +21,29 @@ from dateutil.relativedelta import relativedelta
 from .utils import get_nth_week_number, get_weekday_in_nth_week
 
 
-class UserRoleChoices(models.TextChoices):
-    """Choices for user roles."""
+class RoleChoices(models.TextChoices):
+    """Role choices ."""
 
-    OWNER = "owner"
-    ADMIN = "administrator"
-    MEMBER = "member"
+    MEMBER = "member", _("Member")
+    ADMIN = "administrator", _("Administrator")
+    OWNER = "owner", _("Owner")
+
+    @classmethod
+    def check_administrator_role(cls, role):
+        """Check if a role is administrator."""
+        return role in [cls.ADMIN, cls.OWNER]
+
+    @classmethod
+    def check_owner_role(cls, role):
+        """Check if a role is owner."""
+        return role == cls.OWNER
 
 
-class GroupRoleChoices(models.TextChoices):
-    """Choices for group roles."""
-
-    ADMIN = "administrator"
-    MEMBER = "member"
+# Group roles do not have the "owner" role
+GroupRoleChoices = models.TextChoices(
+    "GroupRoleChoices",
+    [(rc.name, rc.value) for rc in RoleChoices if rc.value != RoleChoices.OWNER],
+)
 
 
 class BaseModel(models.Model):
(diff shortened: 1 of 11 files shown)


src/magnify/apps/core/serializers/rooms.py
```diff
@@ -116,29 +116,37 @@ def to_representation(self, instance):
 
         if not request:
             return output
+
         user = request.user
+        role = instance.get_role(user)
+        is_admin = role in [models.UserRoleChoices.OWNER, models.UserRoleChoices.ADMIN]
```
carofun, line 122:
```suggestion
        role = instance.get_role(user)
        is_admin = role in [models.UserRoleChoices.OWNER, models.UserRoleChoices.ADMIN]
```
don't see the point here, when we have a is_administrator method that does exactly the same job

 return self.get_role(user) in [UserRoleChoices.ADMIN, UserRoleChoices.OWNER]

src/magnify/apps/core/models.py
```diff
@@ -235,39 +235,49 @@ def jitsi_name(self):
...
-    def has_access(self, user):
-        """Check if a user has access to the room."""
-        return self.user_accesses.filter(
-            user=user
-        ).exists() or self.group_accesses.filter(
-            Q(group__members=user) | Q(group__administrators=user)
-        )
+        Users carrying the "owner" role are considered as administrators a fortiori.
+        """
+        return self.get_role(user) in [UserRoleChoices.ADMIN, UserRoleChoices.OWNER]
```
carofun, line 276:
```suggestion
        return self.get_role(user) in [UserRoleChoices.ADMIN, UserRoleChoices.OWNER]
```
if tomorrow GroupRoleChoices.ADMIN !== UserRoleChoices.ADMIN will this still be correct?

src/magnify/apps/core/models.py
```diff
@@ -235,39 +235,49 @@ def jitsi_name(self):
...
+            # pylint: disable=no-member
+            role = self.user_accesses.get(user=user).role
+        except self.user_accesses.model.DoesNotExist:
+            role = None
+
+        if direct_only:
+            return role
+
+        if not (role and role != UserRoleChoices.MEMBER):
+            for access in self.group_accesses.filter(
```
carofun, line 259:
```suggestion
       if direct_only or role not in (None, UserRoleChoices.MEMBER): :
            return role

        for access in self.group_accesses.filter(
```
I find that easier to read

sampaccoud: Good idea! Very elegant :pray:

src/magnify/apps/core/models.py
sampaccoud: Clearly, the whole concept relies on the fact that the member and administrator roles are common in group and user accesses. This is enforced in the definition of choices.

If this was modified, some tests would immediately fail so this could not be done inadvertently.

src/magnify/apps/core/serializers/rooms.py
sampaccoud: True... The point here was to not incur a new db query.

carofun: then maybe you are missing a static method used by the model as well so you don't repeat the logic of what it means to be an admin

src/magnify/apps/core/models.py
carofun: maybe what disturbs me, is by reading the code, without further reading, I could easily think that I can only be an admin if I am an admin from my user role as it is what is written.
We could maybe just use another pair of variables that aren't attached to user or group but only target the label of admin? such as RoleChoices.ADMIN and RoleChoices.Owner and where UserRoleChoices.ADMIN will be set using RoleChoices.ADMIN and the same for GroupRoleChoices
wdyt?

src/magnify/apps/core/serializers/rooms.py
sampaccoud: Or... rename the variable here to "is_admin_or_owner"? Because is_admin when in fact I'm trying to extend to owner is not clear...

carofun (approved):

src/magnify/apps/core/serializers/rooms.py
sampaccoud: I have proposed a refactoring that should please you: https://github.com/openfun/jitsi-magnify/pull/135/commits/b8d2a013555a7b1a4fcbada65737d7c7b1f5380a

src/magnify/apps/core/models.py
sampaccoud: Same. The refactoring in https://github.com/openfun/jitsi-magnify/pull/135/commits/b8d2a013555a7b1a4fcbada65737d7c7b1f5380a should please you.

src/magnify/apps/core/models.py
sampaccoud: fix: https://github.com/openfun/jitsi-magnify/pull/135/commits/b34bd7e9a6ab2c59ee54bb7f08bec385d4500037

carofun (approved):

carofun (approved):

PR #140, opened 2022-12-30
Make meeting recurrence timezone and dst aware

sampaccoud:
## Purpose

We built meetings around 2 separate date and time fields. This was a naive approach that does not work in an international environment with timeones and daily saving times (DST).
 
## Proposal

We refactor meetings to fully support timezones and DST and update tests to make sure it now works as intended. 

We decided to store datetimes in UTC and add a `timezone` field on the meeting to record the timezone of the meeting organizer. This is the timezone in which weekdays are derived. The day of meeting occurences may vary for other participants depending on their own timezone and DST. 

Conversion between UTC and each user's timezone for display in their calendar is left to the frontend so that our API always talks UTC.


kernicPanel (approved):
LGTM :white_check_mark:

PR #144, opened 2023-01-11
Build frontend in container

sampaccoud:
## Purpose

We want to package the frontend inside the Docker image so that Magnify can be served by one simple pod and configured via environment variables.

## Proposal

- Refactor the docker-compose project in order to serve a production stack: nginx+magnify production image without any volume mount so we are really in the conditions of production
- Add [Django Whitenoise](https://github.com/evansd/whitenoise) as a backend dependency to serve static files
- Build the frontend in the Docker production image and collect statics in the container under `/data/static` so Whitenoise can serve them
- Add an API endpoint for frontend configuration so that the Keycloak URL, the site domain, etc. can be configured without rebuilding the Docker image.
- I also took the opportunity to do some clean-up in the `setup.cfg` file information and unpinned the dependencies so Magnify can be installed as a third party app in another project.


sampaccoud:
@lunika I think you should read to give your expertise on the nginx config and also to discuss the approach I took to split urls in nginx and avoid serving files via a Django view like it was done in Marsha. I think it should be more efficient and simpler but hope the nginx config is not too "peculiar"...

.dockerignore
```diff
@@ -27,7 +27,7 @@ docs
 
 # Assets
 data
-src/frontend
+src/frontend/node_modules
```
lunika, line 30: I would suggest you to use `**/node_modules` instead.

docker/files/etc/nginx/conf.d/default.conf
```diff
@@ -4,12 +4,39 @@ server {
...
+    location / {
+        rewrite ^(.*)$ /static/frontend/index.html last;
+    }
+
+    location /index.html {
+        try_files $uri @proxy;
+    }
+
+    # Frontend static files (file with any extension)
+    location ~ ^/assets {
```
lunika, line 24: I'm not sure to understand this rule and why you need it. Is it related with how the front application is using static in its code ?

sampaccoud: the front application looks for its static files in `/assets` but the frontend assets are stored in the backend static folder in `/static/frontend/` to be served by whitenoise.

.dockerignore
sampaccoud: Good idea :+1:

docker/files/etc/nginx/conf.d/default.conf
lunika: Ok but doing this how the front is using the CDN to use the static ? Where this information is stored and how it is used ?

.dockerignore
sampaccoud: done

docker/files/etc/nginx/conf.d/default.conf
lunika: Yes but the front application must be aware of the CDN url. Otherwise it will fetch the application url and never the CDN one.

If I write in my code `` then in your inspector you will see a request made on `https://jitsi-magnify.com/assets/img/dummy.png` and not on the CDN.
I f the application use the webpack split feature, webpack must be aware of the CDN url, otherwise each chunk will be fetch on the application URL and not on the CDN.

sampaccoud: Ah. But the application itself is hosted behind the CDN.
But anyway we are planning a manifest.json at the root of the project that the frontend will get on startup to know the url of Keycloak... and can also get the url of the CDN if necessary.

lunika: The CDN configuration is made in the terraform of the project ? To see how it is handle.

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+- Required: No
+- Default: `"pass"`
+
+#### DB_PORT
+
+Port used by Magnify to connect to the database.
+
+- Type: number
+- Required: No
+- Default: `5432"`
```
kernicPanel, line 328:
```suggestion
- Default: `5432`
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+Keycloak realm to which Magnify should send users to authenticate.
+
+- Type: String
+- Required: No
+- Default: `magnify`
+- Example: `my-magnify`
+
+#### KEYCLOAK_CLIENT_ID
+
+Keycloak client with which users should be authenticated in order to consumes Magnify.
```
kernicPanel, line 351:
```suggestion
Keycloak client with which users should be authenticated in order to consume Magnify.
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+
+- Type: number
+- Required: No
+- Default: `5432"`
+
+### Keycloak-related settings
+
+#### KEYCLOAK_URL
+
+URL of the Keycloak instance to which Magnify should send users to authenticate.
```
kernicPanel, line 334:
```suggestion
Keycloak URL to which Magnify should send users to authenticate.
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+### Crowdin API access related settings
+
+These settings are only used for build purposes in order to push/pull translations
+to/from Crowdin.
+
+#### CROWDIN_API_KEY
+
+The API access key provided by Crowdin. Only managers can have access to this key.
+
+- Type: string
```
kernicPanel, line 377:
```suggestion
- Type: String
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+- Required: No
+- Default: None
+
+#### CROWDIN_PROJECT_ID
+
+The unique project identifier provided by Crowdin. Only managers can have access to this setting.
+
+- Type: string
```
kernicPanel, line 385:
```suggestion
- Type: String
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+- Required: No
+- Default: None
+
+#### CROWDIN_BASE_PATH
+
+The path from where the `crowdin-cli` will work, this path should point to the `src` directory of magnify.
+
+By default its value is `/app/src` and while you are using the container configured in our `docker-compose` file you have no need to change it.
+
+- Type: string
```
kernicPanel, line 395:
```suggestion
- Type: String
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+- Required: None
+- Default: /app/src
```
kernicPanel, line 397:
```suggestion
- Default: `/app/src`
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+#### DJANGO_ALLOWED_HOSTS
+
+A string of comma separated domains that Django should accept and answer to.
+
+See [Django documentation][allowed-hosts] for more details about this setting.
+
+- Type: List as comma separated string
+- Required: Yes
+- Default: [] (all originating domains are rejected)
+- Example: "example1.com, example2.com"
```
kernicPanel, line 26:
```suggestion
- Example: `example1.com, example2.com`
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+
+### Database-related settings
+
+#### DB_ENGINE
+
+Full path to the Python class of the backend used by Magnify to connect to the database.
+
+- Type: string
+- Required: Yes
+- Default: `"django.db.backends.postgresql_psycopg2"` (Postgresql)
```
kernicPanel, line 287:
```suggestion
- Default: `django.db.backends.postgresql_psycopg2` (Postgresql)
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+
+#### DB_HOST
+
+Address for the database used by Magnify.
+
+- Type: string
+- Required: No
+- Default: `"postgresql"`
```
kernicPanel, line 295:
```suggestion
- Default: `postgresql`
```

docs/env.md
```diff
@@ -0,0 +1,407 @@
...
+- Default: `"magnify"`
+
+#### DB_PASSWORD
+
+Password corresponding to the user specified in `DB_USER` and used by Magnify to
+connect to the database.
+
+- Type: string
+- Required: No
+- Default: `"pass"`
```
kernicPanel, line 320:
```suggestion
- Default: `pass`
```

kernicPanel (approved):
Mostly a documentation polishing review…

sampaccoud:
@kernicPanel @lunika thank you for your reviews :pray: Let's merge it!

PR #145, opened 2023-01-12
♻️(frontend) move public assets

PanchoutNathan:
## Purpose

For the production we need to put the public files in an assets folder

## Proposal

- [x] create public assets folder and move files inside


jbpenrath (approved):
LGTM ✅

PR #149, opened 2023-01-13
✨(users) add timezone field to user API

sampaccoud:
## Purpose

We want to expose the `timezone` field to the frontend.

## Proposal

- [x] Add the `timezone` field to the serializer
- [x] Fix and complement tests to take into account the new field


tests/apps/core/test_core_api_users.py
```diff
@@ -110,6 +111,7 @@ def test_api_users_list_authenticated_by_username(self):
                 "id": str(user1.id),
                 "language": user1.language,
                 "name": user1.name,
+                "timezone": user1.timezone.key,
```
lunika, line 114: I think you should open an issue to remove the usage of variable in the assertion and replace them with their values like we do in other projects. WDYT ?

lunika (approved):

tests/apps/core/test_core_api_users.py
sampaccoud: Not sure why. I see value in random testing here no?

PR #153, opened 2023-01-23
🐛(frontend) fix and add many points for the first release 

PanchoutNathan:
## Purpose

A few small bugs or annoying points were listed before the start of production.

## Proposal

- [x] Fix the menu dropdown width 
- [x] Added a modal to display an error when the user makes a request while not logged in
- [x] Display a message in case of error when retrieving room settings
- [x] Autofocus for the add room input
- [x] Add generic error messages

### DIFF
--- a/src/frontend/magnify/apps/magnify-site/src/views/rooms/settings/index.tsx
+++ b/src/frontend/magnify/apps/magnify-site/src/views/rooms/settings/index.tsx
@@ -8,12 +8,13 @@ import {
   useTranslations,
 } from '@openfun/magnify-components';
 import { useQuery, useQueryClient } from '@tanstack/react-query';
-import { Box, Spinner } from 'grommet';
+import { Box, Notification, Spinner } from 'grommet';
 import * as React from 'react';
 import { useEffect } from 'react';
 import { defineMessages } from 'react-intl';
 import { useNavigate, useParams } from 'react-router-dom';
 import { DefaultPage } from '../../../components/DefaultPage';
+import { commonMessages } from '../../../i18n/commonMessages';
 import { RoomsPath } from '../../../utils/routes/rooms';
 
 export const roomSettingsMessages = defineMessages({
@@ -29,7 +30,11 @@ export function RoomSettingsView() {
   const { id } = useParams();
   const queryClient = useQueryClient();
   const navigate = useNavigate();
-  const { data: room, isLoading } = useQuery([MagnifyQueryKeys.ROOM, id], () => {
+  const {
+    data: room,
+    isLoading,
+    error,
+  } = useQuery([MagnifyQueryKeys.ROOM, id], () => {
     return RoomsRepository.get(id);
   });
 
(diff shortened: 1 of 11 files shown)


src/frontend/magnify/packages/components/src/components/rooms/RoomConfig/RoomConfig.tsx
```diff
@@ -122,6 +126,9 @@ export const RoomConfig = ({ room }: RoomConfigProps) => {
           return newRooms;
         });
       },
+      onError: (error) => {
+        modals.handleNotLoggedRequest(error);
```
NathanVss, line 130: I would maybe have used an abstraction between errors and modals. Because: what if some day it isn't modals that are used for errors ? But Toast, etc 

Like this: 

```
// errors/index.ts

function useErrors() {
	const modals = useModals();
	return {
		onError: (error) => {
			modals.onError(error);
			// Do other tasks such as report on Sentry, etc .. 
		}
	}
}
```

src/frontend/magnify/packages/components/src/components/rooms/RegisterRoomForm/RegisterRoomForm.tsx
```diff
@@ -57,6 +59,9 @@ export const RegisterRoomForm = ({ onSuccess }: RegisterRoomFormProps) => {
       });
       onSuccess(newRoom);
     },
+    onError: (error) => {
+      modals.handleNotLoggedRequest(error);
```
NathanVss, line 63: Why did you named it `handleNotLoggedRequest` and not just `handleErrors` ? How are you sure that it's because the user is not logged and not anything else ?

src/frontend/magnify/apps/magnify-site/public/assets/config_exemple.json
```diff
@@ -0,0 +1,8 @@
+{
```
NathanVss, line 1: 🫶

src/frontend/magnify/apps/magnify-site/public/assets/config_exemple.json
```diff
@@ -0,0 +1,8 @@
...
+  "api_url": "http://localhost:7071/api/",
```
NathanVss, line 2: Where is this file used in this commit ?

src/frontend/magnify/apps/magnify-site/src/views/rooms/settings/index.tsx
```diff
@@ -29,7 +30,11 @@ export function RoomSettingsView() {
   const { id } = useParams();
   const queryClient = useQueryClient();
   const navigate = useNavigate();
-  const { data: room, isLoading } = useQuery([MagnifyQueryKeys.ROOM, id], () => {
+  const {
+    data: room,
+    isLoading,
+    error,
+  } = useQuery([MagnifyQueryKeys.ROOM, id], () => {
```
NathanVss, line 37: Why are you not using your generic error modal ?

NathanVss (approved):

PR #154, opened 2023-01-24
🐛(frontend) fix handle error register room form

PanchoutNathan:
## Purpose

The error modal appears even when it's an error handled by react. Like for example a room that already exists

## Proposal

Show modal only when it's not a known error


jbpenrath (approved):
LGTM ✅

PR #155, opened 2023-01-25
♻️(user) make "email" field optional on User

sampaccoud:
## Purpose

We would like to get the user email from the JWT token but should not fail if it is not present.

## Proposal

- [x] Make `email` optional on `User` model
- [x] Add tests to secure the DelegatedJWT authentication backend


tests/apps/core/test_core_authentication.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+    def test_authentication_delegated_jwt_unknown_user(self):
+        """An unknown user should be created by the backend."""
+        jwt_token = AccessToken()
+        jwt_token["sub"] = uuid4()
+        jwt_token["username"] = "dave"
+
+        authenticated_user = DelegatedJWTAuthentication().get_user(jwt_token)
+        new_user = User.objects.get()
+        self.assertEqual(new_user, authenticated_user)
```
jbpenrath, line 43: What do you think to ensute that username has been persisted?

jbpenrath (approved):
LGTM ✅

tests/apps/core/test_core_authentication.py
sampaccoud: Yep. Should have been done :+1:

PR #156, opened 2023-01-26
🔧(magnify-site) disable treeshaking for production build

jbpenrath:
## Purpose

Currently the production build does not work as rollup treeshake remove from bundle the ThemeContext module provided by grommet. For now, the only workaround, we found is to disable treeshake.

## Proposal

- [x] Disable rollup treeshake to build magnify-site for production


sampaccoud (approved):

sampaccoud:
@jbpenrath  Maybe open an issue here too?

NathanVss (approved):
Snif 🥲

PR #158, opened 2023-01-27
♻️(meetings) remove dashes from Jitsi name for meetings

sampaccoud:
## Purpose

We want Magnify to work with SIP gateways which apparently forbid the "-" character in their names. We can remove it as it will not make the Jitsi name less unique.

## Proposal

Drop the "-" character from Jitsi names for meetings and fix related tests.


tests/apps/core/test_core_api_meetings.py
```diff
@@ -212,6 +212,7 @@ def test_api_meetings_list_authenticated_format(self, mock_token):
                 },
             ],
         )
+        meeting_name = str(meeting.id).replace("-", "")
```
kernicPanel, line 215: Why don't use `meeting.jitsi_name` here, and in the other tests ?

sampaccoud: Good idea. I can do that if I add a test in `test_models_meetings.py` to secure the format of this computed.attribute. Let me do that!

kernicPanel (approved):

PR #162, opened 2023-01-30
✨(users) allow differentiating device users from human users

sampaccoud:
## Purpose

Some users are in fact connected to devices in a room. We want to be able to identify and differentiate these special users in the code. 

## Proposal

The idea is to identify them by the fact that they should use a different client in Keycloak and therefore have a different audience set in their token.


src/magnify/apps/core/authentication.py
```diff
@@ -26,9 +26,17 @@ def get_user(self, validated_token):
...
-        # We trust the provider for user authentication
-        defaults.update({"password": "!", "is_active": True})
 
+        # Check audience for known device client
+        audience = validated_token.get("aud")
+        is_device = (
+            audience in getattr(settings, "JWT_USER_DEVICE_AUDIENCES", [])
+            if audience
+            else False
+        )
```
kernicPanel, line 36: I feel that 
```is_device = validated_token.get("aud") in getattr(settings, "JWT_USER_DEVICE_AUDIENCES", [])``` would be more clear and readable.

kernicPanel (approved):
Just some nitpicking ;)

src/magnify/apps/core/authentication.py
sampaccoud: KISS :kissing_heart:

PR #168, opened 2023-02-06
💚(factories) fix randomly failing tests due to fields uniqueness

sampaccoud:
## Purpose

Usernames and room slugs are unique in database but Faker can sometimes generate 2 objects with the same value for these fields.

## Proposal

 To prevent this we introduce a `UniqueFaker` util.


kernicPanel (approved):
Didn't know about this :)

PR #169, opened 2023-02-10
✨(frontend) add LANGUAGE_CODE inside front config

PanchoutNathan:
## Purpose

Today the default language is hard-coded into the code. However, the back offers the LANGUAGE_CODE variable which is configurable. We therefore add this possibility

## Proposal

- [x] add LANGUAGE_CODE inside window.config and use it as default value


src/frontend/magnify/packages/components/src/utils/settings.ts
```diff
@@ -1,3 +1,7 @@
+import { MagnifyLocales } from './constants';
+
 export const KEYCLOAK_TOKEN_VALIDITY = 30;
 
 export const DEFAULT_JITSI_DOMAIN = 'meeting.education';
+
+export const getDefaultLocale = (): string => window.config.LANGUAGE_CODE ?? MagnifyLocales.EN;
```
jbpenrath, line 7: According to the Typescript type LANGUAGE_CODE cannot be `null`

src/frontend/magnify/packages/components/src/i18n/TranslationProvider/TranslationsProvider.tsx
```diff
@@ -43,8 +44,8 @@ const LocaleContext = React.createContext<Maybe<LocaleContextInterface>>(undefin
 
 export const TranslationProvider = ({
   children,
-  locale = MagnifyLocales.EN,
-  defaultLocale = MagnifyLocales.EN,
+  locale = window.config.LANGUAGE_CODE ?? getDefaultLocale(),
+  defaultLocale = window.config.LANGUAGE_CODE ?? getDefaultLocale(),
```
jbpenrath, line 48: You could use `getDefaultLocale` directly. The test you are making here is also did within this method.

jbpenrath (approved):
Add an entry into the CHANGELOG please

PR #173, opened 2023-03-16
🐛(core) base room.jitsi_name only on room.slug

jbpenrath:
## Purpose

Currently, the room.jitsi_name bound into the jwt token generated to be authenticated on the jitsi instance is not the same as the real jitsi room the user tries to access. It appears this difference can prevent a user to enter into a jitsi meeting according the jitsi server configuration. As the slug of a room is unique, we can safely use only this field to build the room.jitsi_name.

In this way real room name and room name bound to the jwt token are the same.

## Proposal

- [x] Build room.jitsi_name property only from its slug
- [x] Update test suites


sampaccoud:
Closed in favor of PR https://github.com/openfun/jitsi-magnify/pull/174

PR #174, opened 2023-03-28
Use UUID (with no dashes) as room name in Jitsi

sampaccoud:
## Purpose

Some jitsi instances using magnify are banning special characters from their room names. 

## Proposal

We achieve compatibility if we use simple uuids with no dashes as Jitsi room names.


tests/apps/core/test_core_api_rooms.py
```diff
@@ -166,6 +166,25 @@ def test_api_rooms_retrieve_anonymous_private_pk(self):
...
+                "id": str(room.id),
+                "is_administrable": False,
+                "is_public": False,
+                "name": room.name,
+                "slug": room.slug,
+            },
+        )
+
     def test_api_rooms_retrieve_anonymous_private_slug(self):
         """It sbhould be possible to get a room by its slug."""
```
jbpenrath:
```suggestion
        """It should be possible to get a room by its slug."""
```

jbpenrath (approved):
LGTM ✅

PR #176, opened 2023-04-04
Allow setting default language

sampaccoud:
## Purpose

We want to allow setting the default language via environment variables.

## Proposal

Add a Django Configuration value for the `LANGUAGE_CODE` setting and update the docs.


jbpenrath (approved):
LGTM ✅

PR #180, opened 2023-04-14
🔧(demo) set CSRF_TRUSTED_ORIGINS for the demo

sampaccoud:
## Purpose

The demo stack was broken because it uses Production settings but we did not set CSRF trusted origins.

## Proposal

Set `CSRF_TRUSTED_ORIGINS` in the demo environment.


jbpenrath (approved):
LGTM ✅

PR #181, opened 2023-04-14
✨(frontend) allow hiding the register link

sampaccoud:
## Purpose

Some identity federations to which Magnify will be connected do not allow registration. We need to hide the "register" link in this case.

## Proposal

- [x] In backend, add an environment variable setting to allow hiding the `register` link: `SHOW_REGISTER_LINK`
- [] In frontend, hide the `register` link if the setting returned by the backend asks so
- [] In frontend, hide the list of rooms when the user is not logged-in and the `register` link is disabled


.circleci/config.yml
```diff
@@ -349,6 +351,8 @@ jobs:
           DJANGO_CONFIGURATION: Test
           DJANGO_SECRET_KEY: ThisIsAnExampleKeyForTestPurposeOnly
           JWT_JITSI_SECRET_KEY: ThisIsAnExampleKeyForDevPurposeOnly
+          MAGNIFY_API_URL: http://localhost:8070/api
+          KEYCLOAK_URL: http://localhost:8080
```
jbpenrath, line 355: Should be gather within an alias ?

src/frontend/magnify/apps/magnify-site/public/config.json
```diff
@@ -4,5 +4,6 @@
   "KEYCLOAK_CLIENT_ID": "magnify-front",
   "KEYCLOAK_EXPIRATION_SECONDS": 1800,
   "KEYCLOAK_REALM": "magnify",
+  "MAGNIFY_SHOW_REGISTER_LINK": false,
```
jbpenrath, line 7: Should be set to `true` for dev purpose nope ?

src/frontend/magnify/packages/components/src/components/rooms/MyRooms/MyRooms.tsx
```diff
@@ -38,63 +38,69 @@ export const MyRooms = ({ baseJitsiUrl, rooms = [], ...props }: MyRoomsProps) =>
...
-        <div>{isLog && <RegisterRoom />}</div>
-      </Box>
-      {isLog && (
-        <>
-          {props.isLoading && (
-            <Box align={'center'} height={'100px'} justify={'center'}>
-              <Spinner />
-            </Box>
+    <>
+      {(isLog || showClaimingRoom) && (
```
jbpenrath, line 45: In order to return early we could write something like that :

```ts
if (!showClaimingRoom) return null;

return (<CustomCard>...</CustomCard)
```

jbpenrath (approved):
Some nitpicks, otherwise LGTM ✅

jbpenrath:
An entry in the CHANGELOG is missing

src/frontend/magnify/apps/magnify-site/public/config.json
PanchoutNathan: Yes

src/frontend/magnify/packages/components/src/components/rooms/MyRooms/MyRooms.tsx
PanchoutNathan: Yes why not :)

PR #183, opened 2023-04-14
(frontend) use 'back' link in Room settings page

### LINKED ISSUE #175: Add a close icon to the room settings page
## Feature Request

**Problem**:
Currently, when a user wants to leave the room's settings page, he/she has to click on the browser back button or on the "Salles" link in the breadcrumb. While this works well, adding a more explicit way to get back could improve the UX.

**Solution**
I propose to add a cross btn on the right side of the page. The screenshots were done with a `grommet` icon as it is already installed. 

Screenshot from 2023-04-03 17-14-27
Screenshot from 2023-04-03 17-15-29

**Do you want to work on it through a Pull Request?**
Sure !

SofienM:
Resolves #175 

The only way to leave the room settings page was to click on the Breadcrumb. 
To improve this UX aspect we change it into a proper back link. 
We use the same link as in User settings page to be coherent accross the app.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,3 +16,4 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 - Use the backend's jitsi name to match the JWT token and set the subject to
   the actual room name.
 - Fix the css of scrolling of the `SimpleLayout` component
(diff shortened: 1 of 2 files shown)


PanchoutNathan (approved):
Change your commit title and LGTM :)

PR #188, opened 2023-05-09
⬆️(docker compose) update docker compose

RaoufCherif:
Update to latest docker compose instead of old docker-compose

## Purpose

Use latest version of docker compose (the compose plugin) instead of old docker-compose tool.

## Proposal

- I checked all references to `docker-compose` and replaced it when needed. 
- I recreated the project with `make bootstrap` and everything seems fine


sampaccoud (approved):
Good move, thank you!

Issue #196, opened 2023-05-26
Cannot create rooms when login is firtsname.lastname


ncasnap:
## Bug Report

**Problematic behavior**
When I create a username firstname.lastname I get an error message :

![image](https://github.com/openfun/jitsi-magnify/assets/107538862/f4f4c24e-08eb-47cb-88ac-927345c03def)

Moreover we don't have the icon with initials of user.
![image](https://github.com/openfun/jitsi-magnify/assets/107538862/00562801-c0b2-43b1-9371-aabe273079e3)

**Expected behavior/code**
We should be able to create rooms and see initials

**Steps to Reproduce**
Create user with username settings to firstname.lastname : 
![image](https://github.com/openfun/jitsi-magnify/assets/107538862/f17c3bcc-5da3-4af8-988f-372e551b4a02)

**Environment**
- Jitsi magnify version: Last version

Thanks in advance for your help and feedback.
Regards,
Nathan

PR #198, opened 2023-05-27
♻️(models) allow customizing username regex & accept periods by default

sampaccoud:
## Purpose

Some authentication systems will accept different sets of characters for their usernames. We should allow magnify to adapt. Accepting periods by default seems legit.

## Proposal

- Use a placeholder string for the username validator message to allow customizing it via translation .po files
- Accept period characters in usernames by default
- Allow customizing the characters accepted for the username by setting the `MAGNIFY_USERNAME_REGEX` environment variable
- Update the environment variables documentation to add the new setting
- Configure Keycloak to validate usernames with the same pattern as Django

Fixes https://github.com/openfun/jitsi-magnify/issues/196


tests/apps/core/test_core_models_users.py
```diff
@@ -92,7 +100,22 @@ def test_models_users_username_ascii(self):
...
             [
-                "Username must contain only lower case letters, numbers, underscores and hyphens."
+                "Username must contain only lower case letters, numbers, hyphens, "
+                "periods and underscores."
+            ],
+        )
+
+    @override_settings(USERNAME_REGEX=r"^[a-z0-9_-]+$")
+    def test_models_users_username_override_regex(self):
+        """It should be possible to override the username regex to allow custom characters."""
```
jbpenrath, line 110: I think the doc string does not describe well the behavior tested as you are removing period from the regex rather to allow new chars.

sandbox/settings.py
```diff
@@ -213,6 +213,12 @@ class Base(MagnifyCoreConfigurationMixin, Configuration):
         environ_prefix=None,
     )
 
+    USERNAME_REGEX = values.Value(
+        r"^[a-z0-9_.-]+$",
+        environ_name="MAGNIFY_USERNAME_REGEX",
+        environ_prefix=None,
```
jbpenrath, line 219: I wonder if, by default, Keycloak is not less stricter about username allowed characters. Shouldn't we use the default username keycloak validator as default value otherwise a user could register on Keycloak then raise a ValidationError when requesting api for the first time ?

jbpenrath (approved):
LGTM ✅

sandbox/settings.py
sampaccoud: yes you're right... we need to figure this out then :thinking:

sampaccoud: I fixed keycloak's config in commit https://github.com/openfun/jitsi-magnify/pull/198/commits/bf2251024491cdd4cfa57aa2936926eaab30de17

tests/apps/core/test_core_models_users.py
sampaccoud: You're right! I improved the docstring

jbpenrath (approved):
LGTM ✅

sandbox/settings.py
jbpenrath: It works well! I just don't know how to apply realm changes without resetting my db, but I think cli is able to do that.

docs/env.md
```diff
@@ -260,6 +260,21 @@ an external tool and Magnify creates users on-the-fly based on OIDC tokens recei
...
 
+#### MAGNIFY_USERNAME_REGEX
+
+Defines what characters are accepted on the `username` field of the user model.
+When customizing it, don't forget to update your translations to customize the
+message held by the placeholder string `username_validator_message`.
+
+Make sure your identity provider (Keycloak in the case of our development sandbox)
+is not moe pemissive on usernam characters otherwise a user can be created in the
+identity provider with a username that will then be rejected by Django.
```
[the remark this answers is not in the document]
sampaccoud: :thinking: bread crumb stuck underneath the "r"

Issue #208, opened 2023-06-01
Public room utilization


ncasnap:
## Bug Report

**Problematic behavior**
When I add a public room to my account, it automatically replicate it to every account.

If we have 100 users create 1 public room, every account dashboard  will be quickly overload.

**Expected behavior/code**
When we create a public room, the room is only visible in our account. And then we share the link to join.

**Steps to Reproduce**
1. Create a public room in your account
2. Check on another account, the room will be present.

I stay available if you need more information.
Regards,
Nathan

PR #215, opened 2023-06-13
Hide public rooms from list endpoint

sampaccoud:
## Purpose

Currently, public rooms are listed on every users' dashboard. Although we want every user be able to join a public room, we don't want them to see it in their personal list of rooms.

We want rooms to be private by default (even better make it configurable via settings)

Finally, the Jitsi parameter to start with audio muted does not work.

## Proposal

- [x] Filter out public rooms on rooms' list API endpoint.
- [x] Add a new setting `DEFAULT_IS_PUBLIC` to determine if a room is public or private by default
- [x] Fix typo on the parameter to start with audio muted

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -223,6 +223,19 @@ for the conference.
 - Default: ""
 - Example: `123`
 
+#### MAGNIFY_DEFAULT_ROOM_IS_PUBLIC
+
+Whether rooms are created as public by default. Rooms that are not marked as public, will not
+deliver any Jitsi JWT token to users who are not authenticated or to users who have not been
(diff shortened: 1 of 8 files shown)


docs/env.md
```diff
@@ -223,6 +223,19 @@ for the conference.
 - Default: ""
```
jbpenrath, line 223: Maybe add a entry into the Changelog about this feature

jbpenrath (approved):
LGTM ✅

Moustique82:
link with issu 208
https://github.com/openfun/jitsi-magnify/issues/208

PR #220, opened 2023-06-16
🚑️(frontend) fix MagnifyModal

PanchoutNathan:
## Purpose

A translation error has crept into the MagnifyModal component

### DIFF
--- a/src/frontend/packages/core/src/components/design-system/Modal/index.tsx
+++ b/src/frontend/packages/core/src/components/design-system/Modal/index.tsx
@@ -58,8 +58,7 @@ export const MagnifyModal = ({
               props.validateButtonCallback?.();
             }}
           >
-            props.validateButtonLabel ?? intl.formatMessage(commonMessages.yes)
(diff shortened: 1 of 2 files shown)


src/frontend/packages/core/src/components/design-system/Modal/index.tsx
```diff
@@ -58,8 +58,7 @@ export const MagnifyModal = ({
               props.validateButtonCallback?.();
             }}
           >
-            props.validateButtonLabel ?? intl.formatMessage(commonMessages.yes)
-            {intl.formatMessage(commonMessages.cancel)}
+            {props.validateButtonLabel ?? intl.formatMessage(commonMessages.yes)}
```
sampaccoud, line 61: It means we are missing a test? add a test to secure it?

jbpenrath: +1

jbpenrath (approved):
LGTM ✅

PR #226, opened 2023-06-20
✨(frontend) Make the "Members" section always visible and functional

### LINKED ISSUE #219: The "Members" section of the room settings page should always be visible
# Bug Report

**Problematic behavior**
When a room is marked as public, its list of members is hidden

**Expected behavior/code**
We may want to declare other owners even if the room is public.

**Steps to Reproduce**
1. Navigate to the room settings page
2. Select False on the "public room" switch
3. Notice how the "Members" section disappears

**Possible Solution**
Make the "Members" section always visible and functional.
Secure behavior with a test.

**Additional context/Screenshots**
You should also add padding at the bottom of the page.

PanchoutNathan:
## Purpose

When a room is marked as public, its list of members is hidden

## Proposal

Make the "Members" section always visible and functional.
Secure behavior with a test.

Close #219

### DIFF
--- a/src/frontend/packages/core/src/views/rooms/settings/index.test.tsx
+++ b/src/frontend/packages/core/src/views/rooms/settings/index.test.tsx
@@ -33,7 +33,7 @@ describe('RoomSettingsView', () => {
     screen.getByText('Members');
   });
 
-  it('renders privfate room', async () => {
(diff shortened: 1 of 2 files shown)


jbpenrath (approved):
LGTM ✅

PR #232, opened 2023-07-07
Issue#231  align selects on list of meeting members

### LINKED ISSUE #231: List of meeting members: align Selects
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

Regarding the management of members of a meeting, the owner of a meeting cannot be removed from the meeting. And so it is not necessary to add the menu.
But, the fact of not adding the menu leads to a non-alignment of the Select which defines the role of the owner of the meeting with the Select of the other members.

**Describe the solution you'd like**
For users who do not have menus, in this case the owner of the meeting, instead of the menu, it would be necessary to have an empty Box which serves as a filling in order to balance the offset.

**Describe alternatives you've considered**
N/A

**Discovery, Documentation, Adoption, Migration Strategy**
N/A

**Do you want to work on it through a Pull Request?**
yes!

adriennickson:
Resolve #231, close #231 

## Purpose
Related to the issue #231

## Proposal
The proposed solution is to replace the missing menu with an empty Box component.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,11 @@ build-front:
 	@$(COMPOSE_RUN) -e HOME="/tmp" -w /app/src/frontend node yarn install
 	@$(COMPOSE_RUN) -e HOME="/tmp" -w /app/src/frontend node yarn build
 
+run-front: ## start frontend development server with live reload
+run-front:
+	@$(COMPOSE_RUN) -e HOME="/tmp" -w /app/src/frontend node yarn install
+	@$(COMPOSE_RUN) -e HOME="/tmp" -w /app/src/frontend -p 3200:3200 node yarn dev
+
 test-front: ## Test frontend for each package
 test-front:
 	@$(COMPOSE_RUN) -e HOME="/tmp" -w /app/src/frontend node yarn test
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,12 +58,12 @@ dev =
     ipdb==0.13.13
     ipython==8.14.0
     isort==5.12.0
-    lxml==4.9.2
+    lxml==4.9.3
     mysqlclient==2.2.0
     msgpack==1.0.5
     pylint==2.17.4
     pylint-django==2.5.3
-    pytest==7.3.2
+    pytest==7.4.0
     pytest-cov==4.1.0
     pytest-django==4.5.2
     responses==0.23.1
@@ -77,7 +77,7 @@ sandbox =
     factory-boy==3.2.1
     gunicorn==20.1.0
     psycopg2-binary==2.9.6
-    sentry-sdk==1.25.1
+    sentry-sdk==1.27.0
 
 [options.packages.find]
 where = src
--- a/src/frontend/packages/core/src/components/users/row/base/RowUserBase.test.tsx
+++ b/src/frontend/packages/core/src/components/users/row/base/RowUserBase.test.tsx
@@ -14,6 +14,12 @@ describe('RowUserBase', () => {
     await screen.findByText(randomUser.name);
   });
 
+  it('basic render with empty Box instead of Menu if there is not actions defined', async () => {
+    const randomUser = createRandomUser();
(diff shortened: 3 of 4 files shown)


src/frontend/packages/core/src/components/users/row/base/RowUserBase.test.tsx
```diff
@@ -14,6 +14,14 @@ describe('RowUserBase', () => {
     await screen.findByText(randomUser.name);
   });
 
+  it('basic render with empty Box instead of Menu if there is not actions defined', async () => {
+    const randomUser = createRandomUser();
+    renderWrappedInTestingProvider(<UserRowBase user={randomUser} />);
+    const menuBox = await screen.findByTestId('menuBox')
+    const emptyBox = menuBox.querySelector(".emptyBox")
+    expect(emptyBox).toBeInstanceOf(HTMLElement);
```
PanchoutNathan, line 22: Why not directly search for the presence of the element by also adding a data-testid to it?

```
await screen.findByTestId('emptyBox')
```

src/frontend/packages/core/src/components/users/row/base/RowUserBase.test.tsx
```diff
@@ -17,8 +17,8 @@ describe('RowUserBase', () => {
   it('basic render with empty Box instead of Menu if there is not actions defined', async () => {
```
PanchoutNathan, line 17: Squash this commit with the previous commit.

src/frontend/packages/core/src/components/users/row/base/RowUserBase.test.tsx
adriennickson: Hi @PanchoutNathan thanks for feedback 🙏🏾.
I just pushed updates.

src/frontend/packages/core/src/components/users/row/base/RowUserBase.test.tsx
adriennickson: Done

Makefile
```diff
@@ -197,6 +202,7 @@ superuser: ## Create an admin user with password "admin"
 	@$(COMPOSE) up -d mysql
 	@echo "Wait for services to be up..."
 	@$(WAIT_DB)
+	@$(MANAGE) migrate
```
sampaccoud, line 205: I don't think we want to force migrate when creating a user. Can you remove this change?

adriennickson: Yes, sure.
Done

setup.cfg
```diff
@@ -58,12 +58,12 @@ dev =
     ipdb==0.13.13
```
PanchoutNathan, line 58: Put this in a separate commit with the title: ``⬆️(dependencies) update python dependencies``

Makefile
```diff
@@ -120,6 +120,11 @@ build-front:
 	@$(COMPOSE_RUN) -e HOME="/tmp" -w /app/src/frontend node yarn install
 	@$(COMPOSE_RUN) -e HOME="/tmp" -w /app/src/frontend node yarn build
 
+run-front: ## start frontend development server with live reload
```
PanchoutNathan, line 123: Put this in separate commit with this `🔧`emoji

src/frontend/packages/core/src/components/users/row/base/RowUserBase.test.tsx
PanchoutNathan: you can remove ```expect(emptyBox).toBeInstanceOf(HTMLElement);```
Because the return type of  
```const emptyBox = await screen.findByTestId('emptyBox');```   
is a ```Promise<HTMLElement>``` and fails the test if it doesn't find it.

setup.cfg
adriennickson: My bad ^^. I squashed my commits with the boot one.

Done

PanchoutNathan (approved):
Resolve all conversations and LGTM (Let's go to the moon) !

